PR #11, opened 2024-07-24
Added a function that returns the intersection of two lists

Liya-F:


utilities/helper-functions.metta
```diff
@@ -236,7 +236,30 @@
...
+(= (member $x $y)
+    (case $y
+        (
+            ( (:: $h $t) (if (== $x $h) True (member $x $t)))
+            ( () False)
+        )
+    )
+)
+
+; Function to return the intersection of two lists
```
Yagth, line 249: It's also best if you test the function whether it works with the tree data structure I gave you yesterday and let me know.

Yagth: I meant, the intersection function should work with list of tree nodes too

EyobAshenaki: Isn't this implemented already? @Yagth

Yagth: @EyobAshenaki Yeah I just saw it. But I don't think it works for a list of tree nodes. Does it?

EyobAshenaki: Yes it doesn't. Maybe it can be modified to accommodate that instead of doing it from scratch.

PR #22, opened 2024-07-29
[ENH] Added TreeNode Equality Checker (Helper Function)

abdulmunimjemal:
A helper function was added to check whether two TreeNode atoms are equal. 
It compares both their value and the constraint's truth value.


EyobAshenaki (changes requested):

EyobAshenaki:
This function is already implemented

PR #25, opened 2024-07-30
[ENH] Implemented  computeGrandchildGuardset function

abdulmunimjemal:
- Added TreeList Difference (setDifference) finding function, which will help compute the setDifference between the grandchild's guard set and result set.
-  Implemented GrandchildGuardset updating function

### DIFF
--- a/utilities/tree-ds.metta
+++ b/utilities/tree-ds.metta
@@ -209,6 +209,32 @@
     )
 )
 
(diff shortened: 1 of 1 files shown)


EyobAshenaki (changes requested):
Re-implement these functions with the new List and Tree data structures in mind

PR #26, opened 2024-07-30
feat: delete-OR-subTree + delete-AND-subTree Implemented

Amanuel-1:
This Pull request intruduces the following changes 
-  **allExceptTargetType** : a function that accepts a TreeList and  returns a new TreeList excluding the target NodeType.
- **delete_OR_subTree** : takes a tree as an argument and removes OR subtrees from the list of children.
- **delete_AND_subTree** : takes a tree as an argument and removes AND subtrees from the list of children.

### DIFF
--- a/enf-reducer/enf-reducer-tests.metta
+++ b/enf-reducer/enf-reducer-tests.metta
@@ -1,0 +1,3 @@
+!(import! &self propagate_truth_value)
+
+!(propagateTruthValue (buildTree (OR a (AND b c))) True)
--- a/utilities/utility-tests.metta
+++ b/utilities/utility-tests.metta
@@ -1,0 +1,235 @@
+! (register-module! ../../metta-moses-reduction)
+! (import! &self metta-moses-reduction:types)
+! (import! &self metta-moses-reduction:utilities:expression-helpers)
+! (import! &self metta-moses-reduction:utilities:list-helpers)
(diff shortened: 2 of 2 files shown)


utilities/helper-functions.metta
```diff
@@ -480,3 +480,82 @@
...
+
+
+
+; below are helper functions for DELETE AND and OR subtree handlers 
+
+
+; takes a list of trees  and a nodeType returns a new tree excluding all the trees with the given NodeType (AND | OR | NOT | LITERAL)
+(: all-except-target (-> TreeList NodeType TreeList))
+(= (all-except-target $treeList $nodeType) (
+    case $treeList
```
EyobAshenaki, line 492: try to use pattern matching when ever possible instead of cases

EyobAshenaki (changes requested):
Re-implement these functions with the new List and Tree data structures in mind

utilities/helper-functions.metta
Amanuel-1: I have reimplemented the functions with the new tree structure and less usage of cases.

Amanuel-1:
> Re-implement these functions with the new List and Tree data structures in mind

EyobAshenaki:
can you change the folder structure so that it fits the current state of the code base

PR #28, opened 2024-07-31
added update-tree which removes a child from listOfChildren and…

Liya-F:
… returns the updated tree


EyobAshenaki (commented):
This functionality can be achieved using `replaceChild`

PR #54, opened 2024-08-07
added a function remove duplicates

Nardos-serkalem:
The is-member function checks if an element x is in a list, returning True if found and False otherwise. The remove-duplicates function uses is-member to remove duplicates from a list, ensuring each element is unique.

### DIFF
--- a/utilities/list-helpers.metta
+++ b/utilities/list-helpers.metta
@@ -261,3 +261,22 @@
     (foldLeft $func $x $xs)
 )
 >>>>>>> df941223a71da2045e09dbc32103c58db2d57f76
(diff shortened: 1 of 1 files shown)


EyobAshenaki (commented):
This functions has already been implemented:
- `is-member` is the same as `member` function found in the list-helpers.metta file
- `remove-duplicates` is the same as `removeDuplicates` higher order function with `member` as it's first argument.

For future reference, use "camelCase" instead of "sneak-case" for function names and move any test cases you have to a separate file that ends with `...test.metta` found in the same folder as the function definition found in.

PR #55, opened 2024-08-07
Remove at index and replace at index implementation

BekLG:
implemented  `replaceAtIndex` and `removeAtIndex` functions, and added their test cases.

### DIFF
--- a/utilities/list-helpers.metta
+++ b/utilities/list-helpers.metta
@@ -271,3 +271,18 @@
         (Cons $x (removeAtIndex (- $a 1) $xs))
     )
 )
+
+ ;A Function to replace an element at a given index in a List.
+(: replaceAtIndex (-> Number $element (List $t) (List $t)))
+(= (replaceAtIndex $a $element Nil) (Cons $element Nil))
+
+(= (replaceAtIndex $a $element (Cons $x $xs))
+    (if (== $a 0)
+        (if (== $xs Nil)
+            (Cons $element Nil)
+            (Cons $element  $xs)
+        )
+
+    (Cons $x (replaceAtIndex (- $a 1) $element $xs))
+)
+)
--- a/utilities/list-helpers.metta
+++ b/utilities/list-helpers.metta
@@ -1,21 +1,21 @@
-; Function to find the length of a list
+ ; Function to find the length of a list
 (:length (-> (List $t) Number))
 (= (length Nil) 0)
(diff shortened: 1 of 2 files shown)


utilities/list-helpers.metta
```diff
@@ -260,4 +260,29 @@
...
 )
->>>>>>> df941223a71da2045e09dbc32103c58db2d57f76
+
+ ;Function to remove an element at a given index in a List.
+(: removeAtIndex (-> Number (List $t) (List $t)))
+(= (removeAtIndex $a Nil) Nil)
+
+(= (removeAtIndex $a (Cons $x $xs))
+    (if (== $a 0)
+        (removeAtIndex (- $a 1) $xs)
```
EyobAshenaki, line 270: The `removeAtIndex (- $a 1)` is unnecessary if index is 0

utilities/list-helpers.metta
```diff
@@ -260,4 +260,29 @@
...
+        (Cons $x (removeAtIndex (- $a 1) $xs))
+    )
+)
+
+ ;A Function to replace an element at a given index in a List.
+(: replaceAtIndex (-> Number $element (List $t) (List $t)))
+(= (replaceAtIndex $a $element Nil) (Cons $element Nil))
```
EyobAshenaki, line 277: If the list is Nil it should return Nil

utilities/list-helpers.metta
```diff
@@ -260,4 +260,29 @@
...
+
+(= (replaceAtIndex $a $element (Cons $x $xs))
+    (if (== $a 0)
+        (if (== $xs Nil)
```
EyobAshenaki, line 281: The if clause is unnecessary here it should just return (Cons $element $xs) if the index is 0

utilities/list-helpers.metta
```diff
@@ -260,4 +260,29 @@
...
+
+(= (removeAtIndex $a (Cons $x $xs))
+    (if (== $a 0)
+        (removeAtIndex (- $a 1) $xs)
+        (Cons $x (removeAtIndex (- $a 1) $xs))
+    )
+)
+
+ ;A Function to replace an element at a given index in a List.
+(: replaceAtIndex (-> Number $element (List $t) (List $t)))
```
EyobAshenaki, line 276: the type of the element to should be the same as the types of the elements found in the list

EyobAshenaki (approved):

PR #57, opened 2024-08-07
Implementation for commandSetIterator function based on the Python implementation.

samwondim:


EyobAshenaki (commented):
This changes include out dated files. You should implement `commandSetIterator` by creating a branch form the main branch

PR #62, opened 2024-08-09
Refactored Propagate Truth Value algorithm

Yagth:
- Changed logic to use built in functions to reduce complexity
- Fixed truth value arg not being updated on LITERAL nodes


EyobAshenaki (approved):
Every thing seems to be in order

PR #63, opened 2024-08-09
Added the functions orSubTreeIterator and andSubTreeIterator

Liya-F:
- Added the functions orSubTreeIterator and andSubTreeIterator inside the RTEHelpers file
- They are the implementations of the same functions found in the python file


EyobAshenaki (changes requested):
**Issues**:
There are multiple undefined functions in this PR, namely `commandSetIterator`, `orSubTreeElegance`, `orSubTreeIterator`, `andSubTreeElegance` and `andSubTreeIterator`. 

**Requested Changes**:
- Implement the above missing functions 
- Include at least 2 test cases for all implemented functions

Yagth:
@EyobAshenaki Those functions are being implemented by other people, we can test them until they are done. What do you think is better? Should we merge the codes and test them later or should we wait until all of them are done and do the test then merge them?

EyobAshenaki:
The latter case seems like the best choice. It would save us from a headache later on.

Yagth:
But we would still face so e interdependence between the functions. For instance, the `orSubTreeElegance` needs to call the `reduceToElegance` algorithm which can't be complete without the `orSubTreeElegance` itself. What if we could replace the function calls with the appropriate return type of the function stored on a variable, instead of calling the functions we put a mock data on a variable and test whether the function works or not?

EyobAshenaki:
I think that's an acceptable compromise time wise.

Yagth:
> @EyobAshenaki But we would still face the same problem letter on because of the interdependence between the functions. For instance, the `orSubTreeElegance` needs to call the `reduceToElegance` algorithm which can't be complete without the `orSubTreeElegance` itself. What if we could replace the function calls with the appropriate return type of the function stored on a variable, instead of calling the functions we put a mock data on a variable and test whether the function works or not?

@Liya-F can you make sure to use a mock data instead of calling the functions that aren't implemented yet? Refactor the code this way, test it out using one or two test cases and update your PR.

Liya-F:
- I have replaced the unimplemented functions with mock data. Specifically, I have assigned $action with ADVANCE, which will be the value returned by the functions once they are implemented.
- I have commented out the code that calls these functions for future use.
- I have also added test cases for both

enf-reducer/RTEHelpers.metta
```diff
@@ -90,3 +90,69 @@
         )
     )
 )
+
+;;a function which is the reimplementation of the orSubTreeIterator function found in python
+
+(: orSubTreeIterator(-> Tree (List Tree) Tree (List Tree) (List Tree) Signal))
```
EyobAshenaki, line 96: the type definition should return a tuple containing Signal and Tree

enf-reducer/RTEHelpers.metta
```diff
@@ -90,3 +90,69 @@
...
+        ($else ($action $currentNode))
+
+     )
+
+ )
+))
+
+;;a function which is the reimplementation of the andSubTreeIterator function found in python
+
+(: andSubTreeIterator(-> (List Tree) Tree (List Tree) (List Tree) (-> (List Tree) Tree (List Tree) (List Tree) Number Signal)))
```
EyobAshenaki, line 126: the type definition should return a tuple containing Signal and Tree

enf-reducer/RTEHelpers.metta
```diff
@@ -90,3 +90,69 @@
...
+(= (andSubTreeIterator $children $currentNode $handleSet $commandSet) 
+    (andSubTreeIterator $children $currentNode $handleSet $commandSet 0)
+)
+(: andSubTreeIterator(-> (List Tree) Tree (List Tree) (List Tree) Number Signal))
```
EyobAshenaki, line 130: the type definition should return a tuple containing Signal and Tree

EyobAshenaki (changes requested):
fix the type definition for the implemented functions

Liya-F:
> fix the type definition for the implemented functions

I have fixed the type definition for both functions.

EyobAshenaki (approved):

PR #73, opened 2024-08-17
Test runner script

EyobAshenaki:
This PR contains a test runner that executes all tests found in a MeTTa project.

There are few requirements that needs to be satisfied for the test runner to work smoothly are:

- The script should be executed in the root directory of the project so that it can identify all the test files in the project.
- Test file should always end with `...test.metta` for the script to recognize them and execute the tests found in them.
- In the test files `assert..` expressions should be preceded by `!` for the script to recognize and execute them as tests.


Yagth (commented):
@EyobAshenaki I tried to run the test script after making the following change to one of the `enf-reducer-test.metta` file. I added the following line to the end of the file and the script doesn't seem to be working while the metta file works fine when I run it. Is there anything I missed?
```scheme
!(assertEqual True False)
!(assertEqualToResult (True) (False))
```

Yagth (approved):

PR #77, opened 2024-08-19
Feature/and sub tree bug fix

samwondim:


Yagth:
Hey @samwondim . There are changes not merged from reduceToElgance yet. So it's best if we first test the RTE and then merge the code to main. :)

PR #90, opened 2024-08-26
Unit Tests for all functions

YeabsiraNigusse:
almost all functions are tested and exits in different file according to their functionality

- two RTE helper functions(iterator and orSubTreeElegance) didn't return the intended result
- iterator return deterministic result and orSubTreeElegance return incorrect result compared to python


Yagth (commented):
@YeabsiraNigusse I don't see a file that's not changed. I think you are using a formatter that  changed are you sure everything works?

YeabsiraNigusse:
@Yagth, yes everything is working on my branch and i pull changes from the main branch before sending pull request. the number of file changed increased due to commit from WSL terminal which i was not committing changes before. i accidentally commit and push from that terminal and a lot of changes get into the file but technically there is no difference on there repo since i was using the same folder for the two terminals. if there is anything i should test let me know. you can also switch to my branch and check if everything is working.

Yagth (approved):

samwondim (approved):

PR #102, opened 2024-09-18
refactor: gather-junctors

eyuuab:


enf-reducer/gather-junctors.metta
```diff
@@ -1,48 +1,29 @@
-(: gatherJunctors (-> Tree Tree Tree))
-(= (gatherJunctors $currentNode $centerNode)
-      (let* (
-              ($currentNodeType (getNodeType $currentNode))
-              ($centerNodeType (getNodeType $centerNode))
-              ($currentLeft (getChild $currentNode L))
-              ($currentRight (getChild $currentNode R))
-              ((TreeNode (Value $valueName $constraint $currentNodeType) $guardSet $children) $currentNode)
+; a function to merge two tuples (A B C) (D E F) ==> (A B C D E F)
+(= (concat-tuple $xs $ys)
```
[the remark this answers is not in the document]
Yagth: It will also be cool if you can add test cases for this function using the `assertEqual` in a utility file or even in the gatherJucntor file?

enf-reducer/gather-junctors.metta
```diff
@@ -1,48 +1,29 @@
...
-(= (gatherJunctors $currentNode $centerNode)
-      (let* (
-              ($currentNodeType (getNodeType $currentNode))
-              ($centerNodeType (getNodeType $centerNode))
-              ($currentLeft (getChild $currentNode L))
-              ($currentRight (getChild $currentNode R))
-              ((TreeNode (Value $valueName $constraint $currentNodeType) $guardSet $children) $currentNode)
+; a function to merge two tuples (A B C) (D E F) ==> (A B C D E F)
+(= (concat-tuple $xs $ys)
+    (if (== $xs ()) $ys
```
Yagth, line 3: For better readability, I suggest formatting this line as such:
```scheme
(if (== $xs ())
    $ys
    (let* ...
    )
)
```

enf-reducer/gather-junctors.metta
```diff
@@ -1,48 +1,29 @@
...
+        )
+    )
+)
+
+(: gather-junctors (-> $expr $prev $expr))
+(= (gather-junctors $expr $prev)
+    ( case $expr
+        (
+            (($op $a $b)
+                ( if (== $prev ())
```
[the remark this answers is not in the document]
Yagth: I think you can remove this part of the code and it will still work as expected

enf-reducer/gather-junctors.metta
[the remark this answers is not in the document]
Yagth: And I think this will be a very useful helper function, it will be better if moved somewhere else. Also don't be shy to use some symbols as a function name like `++` or something. We already have a function named as `++` right now but you can find some other symbol that will be easy to understand and even defined as an infix operator instead of prefix like so
```scheme
(: ++ (-> Atom Atom Atom))
(= ($xs ++ $ys) ...)
```
This will make it will better readable whenever someone sees the code.

enf-reducer/gather-junctors.metta
```diff
@@ -1,48 +1,29 @@
...
+                    (concat-tuple ($op) (concat-tuple (gather-junctors $a $op) (gather-junctors $b $op)))
+                    (if (== $op $prev)
+                    (concat-tuple (gather-junctors $a $op) (gather-junctors $b $op))
```
Yagth, line 21: Can you indent the two lines for better readability? Like so:
```scheme
(if (== $op $prev)
    (concat-tuple ...)
    ((concat-tuple ..)
)
```

enf-reducer/gather-junctors.metta
```diff
@@ -1,48 +1,29 @@
...
-                          )
-                          NilNode
-                      )
-                  )
-            )
-         )
- )
+            ($symbol ($symbol))
+        )
+    ))
```
Yagth, line 29: These are also better if closed like so:
```scheme
(= (gather-junctors $expr $prev)
      ( case $expr
            (
             ....
             )
       )
)
```

enf-reducer/gather-junctors.metta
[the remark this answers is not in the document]
eyuuab: That condition is required to handle the case where the output is enclosed in double braces, like (( )).

it will be passed with the test case.  $prev will be **( )** for the _root node_

utilities/list-helpers.metta
```diff
@@ -74,6 +74,18 @@
 (= (extend Nil $ys) $ys)
 (= (extend (Cons $x $xs) $ys) (Cons $x (extend $xs $ys)))
 
+; a function to merge two tuples (A B C) (D E F) ==> (A B C D E F)
```
Yagth, line 77: Better if you move this function below the  `++` function. The `++` function is similar to `extend` so it's better if they are together :)

enf-reducer/gather-junctors.metta
```diff
@@ -1,48 +1,18 @@
...
-                      )
-                      (if (== $currentNodeType LITERAL)
-                          (if (== $centerNodeType AND)
-                              (prependToGuardSet $centerNode $currentNode)
-                              (prependChild $centerNode (TreeNode (Value Nil $constraint AND) (Cons $currentNode Nil) $children))
-                          )
-                          NilNode
-                      )
-                  )
+(: gather-junctors (-> $expr $prev $expr))
```
Yagth, line 1: This name is still not `camelCase`

Yagth (changes requested):
Make sure every function and variable names are in camel case

utilities/list-helpers.metta
```diff
@@ -74,6 +74,18 @@
...
+(= (concatTuple $xs $ys)
+   (if (== $xs ())
+       $ys
+       (let* (($head (car-atom $xs)) 
+              ($tail (cdr-atom $xs)) 
+              ($tail-new (concatTuple $tail $ys)))
```
Yagth, line 83: This variable name should also be in camelCase

utilities/list-helpers.metta
Yagth: Can you put a type definition for the `concatTuple` function too?

PR #103, opened 2024-09-19
feat: implemented nAry-gatherJunctors

Amanuel-1:
this branch refactors gather-junctors for nAry operations

### DIFF
--- a/enf-reducer/tests/gather-junctor-test.metta
+++ b/enf-reducer/tests/gather-junctor-test.metta
@@ -1,86 +1,102 @@
 ! (register-module! ../../../metta-moses-reduction)
-! (import! &self metta-moses-reduction:types) 
-! (import! &self metta-moses-reduction:utilities:list-helpers)
-! (import! &self metta-moses-reduction:utilities:tree-helpers)
-! (import! &self metta-moses-reduction:enf-reducer:rte-helpers)
-! (import! &self metta-moses-reduction:enf-reducer:propagate-truth-value)
-! (import! &self metta-moses-reduction:enf-reducer:reduce-to-elegance)
+;; ! (import! &self metta-moses-reduction:types) 
+;; ! (import! &self metta-moses-reduction:utilities:list-helpers)
+;; ! (import! &self metta-moses-reduction:utilities:tree-helpers)
+;; ! (import! &self metta-moses-reduction:enf-reducer:rte-helpers)
+;; ! (import! &self metta-moses-reduction:enf-reducer:propagate-truth-value)
+;; ! (import! &self metta-moses-reduction:enf-reducer:reduce-to-elegance)
 
-! (import! &self metta-moses-reduction:enf-reducer:post-processing)
-! (import! &self metta-moses-reduction:enf-reducer:reduct)
-! (import! &self metta-moses-reduction:enf-reducer:gather-junctors)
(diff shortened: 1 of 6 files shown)


enf-reducer/n-ary-gather-junctors.metta
```diff
@@ -0,0 +1,72 @@
...
+             )
+         (cons-atom $headWithParams $tailWithParams)  
+       )
+   )
+)
+
+ 
+(= (gatherJunctors $expr) (gather-junctors $expr ()))
+
+!(gather-junctors (NOR a b (OR c d (OR e f)) k) ())
```
Yagth, line 65: Remove this line

enf-reducer/n-ary-gather-junctors.metta
```diff
@@ -0,0 +1,72 @@
...
+              ($tailWithParams (map $f $tail $p))
+             
+             )
+         (cons-atom $headWithParams $tailWithParams)  
+       )
+   )
+)
+
+ 
+(= (gatherJunctors $expr) (gather-junctors $expr ()))
```
Yagth, line 63: What is the purpose of this line here?

enf-reducer/n-ary-gather-junctors.metta
```diff
@@ -0,0 +1,72 @@
...
+        (let* (($head (car-atom $xs))  
+               ($tail (cdr-atom $xs)) 
+               ($tail-new (concat-tuple $tail $ys))  
+              )
+          (cons-atom $head $tail-new)  
+        )
+    )
+)
+
+(= (fold $f $tuple $acc)
```
Yagth, line 12: Can you move it to the `Utilities/general-helper-functions.metta` file. That's where we are going to be keeping functions that deal with any atom.

Yagth: It's also best if you can add a type definition for this function. But don't add the type definition for `$f` as `Atom` it's better to leave it without type definition than give the `$f` a type of `Atom`.

The definition could look something like this:
```scheme
(: fold (-> (-> Atom Atom Atom) Atom Atom)
```

This might sound redundant but it's better in the long run.

Yagth: And also for the `foldl` function, it's not common to put the accumulator in the end. Some one might misunderstand the order of the arguments and cause a problem in the long run. So it's better if you can switch their position to be like so:
```scheme
(= (fold $f $acc $tuple) ...)
```

Yagth (changes requested):
I have added some comments and will add more once you complete those. And also make sure to resolve the Merge conflict as well.

enf-reducer/n-ary-gather-junctors.metta
Amanuel-1: sure . on it

enf-reducer/n-ary-gather-junctors.metta
Amanuel-1: the function gather-junctors function takes two arguments -- the expression itself and an accumulator . but when the function is called for the first time we have to pass an empty tuple for the accumulator . I thought this would make the code unreadable when this function is used .

the function below is just a wrapper that can be used to avoid unnecessary passing of empty arguments

PR #105, opened 2024-09-22
refactor intersections.metta func

Nardos-serkalem:
Refactored the intersections function to eliminate case-based logic and replaced it with pattern matching

### DIFF
--- a/utilities/tree-helpers.metta
+++ b/utilities/tree-helpers.metta
@@ -362,3 +362,25 @@
     )
 )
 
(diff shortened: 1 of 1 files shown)


Yagth:
Hey @Nardos-serkalem is the intersection function working properly? The workflow keeps timing out

Nardos-serkalem:
Hey! yes it's working properly on my end.

Nardos-serkalem:
Hey @Yagth yes it's working properly on my end.

PR #109, opened 2024-09-24
Added RTE function

YeabsiraNigusse:
i think the testes are failing because i removed previous RTE implementation from my new branch. if that is a must and if we went to test this new implementation of RTE pass the required testes first we have to finish the new RTE helper functions (transformations) so we can hold this pull request. otherwise if the implementation looks safe we can merge it.


YeabsiraNigusse:
i think the testes are failing because i removed previous RTE implementation from my new branch

Yagth:
I think it's best if we wait for the implementation of the functions to be done first. Then we can merge a working version. Until then let's hold this PR.

PR #112, opened 2024-09-30
Refactor propagateNot function to be N-ary

Getch-27:


enf-reducer/n-ary-propagate-not.metta
```diff
@@ -0,0 +1,39 @@
+(=(AND $x) (cons-atom AND $x))
+(=(OR $x) (cons-atom OR $x))
+
+;; Propagate to childrens for both (propagateNot NOT) and (propagateNot logic)
+(=(propagateChildrens $tuple) (collapse (propagateNot (superpose $tuple))))
```
Yagth, line 5: Add a spacing between `=` and `(` Here. Like so
```scheme
(= (propagateChildren ...
```

enf-reducer/n-ary-propagate-not.metta
```diff
@@ -0,0 +1,39 @@
...
+          (propagateNot NOT $x (cdr-atom $expr))
+       )
+      )
+      (NOT $expr)
+))
+
+(=(propagateNot $expr ()) (propagateNot $expr))
+(=(propagateNot $expr) 
+  (
+    if (== (get-metatype $expr) Expression)
```
Yagth, line 30: Done separate `if from its left bracket

Yagth (changes requested):
Hey @Getch-27. Really great what you did with the `superpose` function to make the `propagateNot` nary. But it would even be better if you do the following...

1. First is that you should use the same function that can handle both the not cases and normal cases. Using `propagateChildren` for logic case and `propagateNotChildren` for the Not case isn't an advisable approach to follow. You can achieve this using `case`. I understand you did that to avoid non determinism and I said best to avoid `case` function last time but you this would be good place to use it.

2. In the `propagateNot` function definitions, it's best if you can avoid using `get-metatype` if possible (If you can avoid using this by utilizing `case`). This makes the code simpler to read and more elegant. This will increase the readability.

3. Add at least one more test case that has more than two children in the boolean expression. i.e. `(AND (NOT (OR A B)) (NOT (AND C D E)) (OR A B) A B)`

PR #114, opened 2024-10-01
chore: refactor applyOrCutToChildren to use foldLeft implementation r…

samwondim:


Yagth (commented):
Looks like the update is causing some issues with the RTE function. Make sure to test it against the previous implementation of `applyOrCutToChildren`

PR #115, opened 2024-10-01
feature: add a `binaryMap` function to replace map functions that tak…

samwondim:


enf-reducer/rte-helpers.metta
```diff
@@ -137,16 +137,10 @@
...
-(= (mapList $f Nil $rs) Nil)
-(= (mapList $f (Cons $x $xs) $rs)
-  (Cons ($f $x $rs) (mapList $f $xs $rs))
-)
-;;A function that applies a function to the elements of the first list and a constant parent.
-(: mapParent (-> $f (List $t) $t (List $tt)))
-(= (mapParent $f Nil $parent) Nil)
-(= (mapParent $f (Cons $x $xs) $parent)
-  (Cons ($f $x $parent) (mapParent $f $xs $parent))
+(: binaryMap (-> Atom (List $t) (List $t) (List $t)))
```
Yagth, line 140: It's best if you avoid using `Atom` as type. You can define the type of the function that will be accepted as `(-> $t $t $t)`.

Yagth: And shouldn't the second argument for the `binaryMap` function should be of type `$t` instead of `(List $t)`?

Yagth: You can then use this `binaryMap` instead of mapParent :)

Yagth (changes requested):
It would also be cool if you can find other similar functions like `mapParent` which are doing the same thing as the `binaryMap` function.

PR #118, opened 2024-10-02
Chore/refactor apply or cut to chldrn

samwondim:


enf-reducer/rte-helpers.metta
```diff
@@ -75,6 +75,20 @@
         $current
     )
 )
+
+(= (foldLeftBinary $f $acc $current Nil) $acc)
```
Yagth, line 79: Best to have the function have a type and also a comment of what it does. Example:
```scheme
(= foldLeftBinary (-> (-> $t $t $t) $t $t (List $t) $t))
```

enf-reducer/rte-helpers.metta
```diff
@@ -75,6 +75,20 @@
...
+(= (foldLeftBinary $f $acc $current (Cons $x $xs))
```
Yagth, line 80: And I noticed you didn't use the `$acc` in your code in the `foldLeftBinary` function? If you don't use it, remove it from the constraint :).

Yagth (changes requested):

Yagth (approved):

PR #120, opened 2024-10-02
binary map implementation and foldl implementation in applyOrCutToChildren 

Amicahh:


Yagth:
@Amicahh this isn't passing the tests defined. Mind checking if you implemented it correct?

PR #123, opened 2024-10-03
implemented the 0-Constraint-Subsumption Transformation

Liya-F:
I have implemented the transformation which is applied on any OR node. It checks if a child of the OR node has no guardSet and no child and if the condition is satisfied it deletes the node. if not, the node is kept. 
- I have used the getGuardSetExp method to check the guardSet of each child
- And the nodeHasChild method to check if the child has any child.


Nahom32:
@Liya-F  please add comments to your implementations to highlight what it does.

enf-reducer/zero-constraint-subsumption.metta
```diff
@@ -0,0 +1,22 @@
+(= (zeroConstraintSubsume $expOriginal $expRecursive)
+(if (== $expRecursive ()) $expOriginal 
+(let*
+(
+    ($head (car-atom $expRecursive))
+    ($tail (cdr-atom $expRecursive))
+    ($headType (get-metatype $head))
+    ($headHasNoChild (not (nodeHasChildExp $head)))
+)
+(if (== $head AND) $expOriginal
```
Yagth, line 10: Isn't possible to use `case` here instead of nesting `if` expressions? It would make the code more readable.

enf-reducer/tests/zero-constraint-subsumption-test.metta
```diff
@@ -0,0 +1,14 @@
+! (register-module! ../../../metta-moses-reduction) 
+! (import! &self metta-moses-reduction:enf-reducer:rte-helpers)
+! (import! &self metta-moses-reduction:enf-reducer:zero-constraint-subsumption)
+ 
+ ! (assertEqualToResult (zeroConstraintSubsume (OR (AND A B) (AND)) (OR (AND A B) (AND))) (()))
+ ! (assertEqualToResult (zeroConstraintSubsume (OR (AND A B) (OR)) (OR (AND A B) (OR))) (()))
+ ! (assertEqualToResult (zeroConstraintSubsume (OR (AND A B) (A)) (OR (AND A B) (A))) ((OR (AND A B) (A))))
```
Yagth, line 7: Why did you write the `(A)` in the `(OR (AND A B) (A))` test case? Do we have a situation where a literal being enclosed within a tuple?

Yagth (changes requested):

Nahom32 (approved):

enf-reducer/zero-constraint-subsumption.metta
Liya-F: yeah that's right, I have corrected it as such

enf-reducer/tests/zero-constraint-subsumption-test.metta
Liya-F: I don't think there is, I was trying all possible combinations. Should I remove it? also, is (OR) a possibility?

Yagth (approved):

PR #128, opened 2024-10-05
test cases for n-ary-gather-junctors function

YeabsiraNigusse:
@Yagth  next time someone complete a function tell them to add test cases also so it won't be a task and it will be easy too unless you want the function to be tested by another person.


Yagth:
> @Yagth  next time someone complete a function tell them to add test cases also so it won't be a task and it will be easy too unless you want the function to be tested by another person.

I always do to be honest. But to make the tests compete, sometimes it's a must that we write some of our own too. Ideally this should be done even before the task is given out. That way whoever is doing the task will know whether it worked or not before creating a PR.

Nahom32 (approved):

Yagth (approved):

PR #130, opened 2024-10-05
feat: cut unnecessary or transformation

BrookFeleke:


Nahom32 (approved):

enf-reducer/cut-unnecessary-or.metta
```diff
@@ -0,0 +1,19 @@
+ ; Cut unnecessary OR, DOESN'T process sub expressions
+ ; To be used with iterator
```
Yagth, line 2: Can you give a bit more information than that if you can. If you can't just let me know and we will add it later :). The comment should ideally be what the function does, possibly with example input output to give the reader a glance of what it does without the need to look into the code :)

enf-reducer/tests/cut-unnecessary-or-test.metta
```diff
@@ -0,0 +1,44 @@
+! (register-module! ../../../metta-moses-reduction)
+! (import! &self metta-moses-reduction:enf-reducer:cut-unnecessary-or)
+
+ ;; Test 1: OR with multiple top-level children, sub-expressions not handled (no transformation)
+! (assertEqual (orCut (OR a (OR b) (OR c (OR f)))) (OR a (OR b) (OR c (OR f))))
```
Yagth, line 5: This test case is invalid. We won't have a consecutive `OR`s as input for this function. This is due to the `gatherJunctors` algorithm which will make sure that any boolean expression has an alternating sequence of `OR`s and `AND`s always. There fore, The above test case is as simple as: `(OR a b c f)`. Make sure to update your test cases as such.

Yagth: Also make sure to add the test case found on the holman paper. And other similar cases where things get complicated. i.e., `(OR (AND (NOT A) B C D)` and `(OR (AND (NOT A) B C D (OR A B))`.

enf-reducer/cut-unnecessary-or.metta
```diff
@@ -0,0 +1,19 @@
...
+            (if (== $type Symbol)
+                $expresion
+                (let* (
+                        ($head (car-atom $expresion))
+                        ($tail (cdr-atom $expresion))
+                    ) (if (== $head OR)
+                    (let* (
+                            ($subhead (car-atom $tail))
+                            ($subtail (cdr-atom $tail))
+                        ) (if (== $subtail ()) $subhead (cons-atom $head $tail)))
```
Yagth, line 16: This isn't a correct implementation as described on the paper. The function should be returning the `guardSet` and the `children` of the grandchild expression instead of returning the whole grandchild itself as you did here. Make sure to update your code as such.

Yagth: And there is no need to use `(cons-atom $head $tail)` here. You can just return the initial `$expression`

Yagth (changes requested):

enf-reducer/cut-unnecessary-or.metta
BrookFeleke: > This isn't a correct implementation as described on the paper. The function should be returning the `guardSet` and the `children` of the grandchild expression instead of returning the whole grandchild itself as you did here. Make sure to update your code as such.

I checked the description, and I am not sure I understand what you mean, maybe in the form of input and expected output.

Yagth: @BrookFeleke In the paper, if you understood the description correctly. i.e., if the `POA` is for instance the `OR` node in the expression `(AND B (OR (AND (NOT D) C)))` the cut unnecessary or transformation will returns the expression `(AND B (NOT D) C)`. But your implementation is returning `(AND B (AND (NOT D) C)))`. Which isn't correct.

PR #131, opened 2024-10-08
implemented the 1-Constraint-Subsumption Transformation

Liya-F:
The POA for this transformation is any non-root `AND` node. If any constraints of the node are already found inside the `$commandSet` the node is removed, if not, it's kept. 
- I have used the `isMember` method from general-helper-functions.

### DIFF
--- a/enf-reducer/one-constraint-subsumption.metta
+++ b/enf-reducer/one-constraint-subsumption.metta
@@ -1,5 +1,6 @@
 ;a function which implements the 1-Constraint-Subsumption Transformation
 (= (oneConstraintSubsume $expOriginal $expRecursive $commandSet) 
+(if (== $expRecursive ()) $expOriginal
 (let* 
   (
     ($head (car-atom $expRecursive))
@@ -10,8 +11,9 @@
     (OR $expOriginal)
     (NOT $expOriginal)
     (AND (oneConstraintSubsume $expOriginal $tail $commandSet)) ;the POA of this transformation is any non-root AND node
-    ($else (if (isMember $head $commandSet) () $expOriginal)) ;if any constraint of the AND node is found inside the commandSet, the node is removed. if not, it is kept.
+    ($else (if (isMember $head $commandSet) () (oneConstraintSubsume $expOriginal $tail $commandSet))) ;if any constraint of the AND node is found inside the commandSet, the node is removed. if not, it is kept.
   )
   )
   )
+)
 )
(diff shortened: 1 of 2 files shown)


enf-reducer/tests/one-constraint-subsumption-test.metta
```diff
@@ -0,0 +1,20 @@
...
+;Test-04:applied on a NOT expression(not affected)
+ ! (assertEqualToResult (oneConstraintSubsume (NOT A) (NOT A) (A B)) ((NOT A)))
+;Test-05:applied on an AND Node
+;;; empty $commandSet(the Root node or any AND node which is not commanded by any nodes)
+ ! (assertEqualToResult (oneConstraintSubsume (AND A (OR (AND C D) E)) (AND A (OR (AND C D) E)) ()) ((AND A (OR (AND C D) E))))
+;;; an AND node whose constraint is already inside the $commandSet
+ ! (assertEqualToResult (oneConstraintSubsume (AND A B C) (AND A B C) (A D E F)) (()))
+;;; an AND node whose constraint isn't inside the $commandSet 
+ ! (assertEqualToResult (oneConstraintSubsume (AND A B C) (AND A B C) (D E F)) ((AND A B C)))
+
```
Yagth, line 20: Can you add the test case found in the holman paper too?

enf-reducer/one-constraint-subsumption.metta
```diff
@@ -0,0 +1,17 @@
...
+(let* 
+  (
+    ($head (car-atom $expRecursive))
+    ($tail (cdr-atom $expRecursive))
+  )
+  (case $head
+  (
+    (OR $expOriginal)
+    (NOT $expOriginal)
+    (AND (oneConstraintSubsume $expOriginal $tail $commandSet)) ;the POA of this transformation is any non-root AND node
```
Yagth, line 12: Doesn't this line of code jump all the `AND` nodes found after the expression, not just the first root node?

Yagth (changes requested):

enf-reducer/tests/one-constraint-subsumption-test.metta
Liya-F: okay, I have added it. But just the POA part.

enf-reducer/one-constraint-subsumption.metta
Liya-F: No, it doesn't I have checked it. Because, I don't think we will have just AND in our expression. It will always be (AND) or (AND A B ...). I have implemented it by first checking if the root node is AND should I change it to that?

Nahom32 (approved):

Yagth (approved):

Issue #143, opened 2024-10-12
Implement `getDominantSet` function for a given AND node expression


YeabsiraNigusse:
**Detailed Description**

This function implementation helps retrieve the dominant set of a given AND node expression. The output of this function will be used as a parameter for the RTE (Reduce To Elegance) function. We call this function every time the RTE function is invoked for each subexpression. The dominant set for a given AND node is the union of all parent AND node guard sets. For more details, refer to the Holman paper.

**Context**

The implementation of this function will simplify the RTE execution process. The dominant set is an important parameter for the RTE transformation. It might be passed as a parameter to the transformation function or used for conditionals. Since Metta doesn’t use pointers, we can’t track the dominant set each time the expression changes, so it is necessary to call this function whenever the RTE function is called. For more insight into the importance of this function, refer to the RTE function implementation. `If you find a better way to get or track the dominant set for the RTE function, this issue remains open for improvements.`

**Possible Implementation**

This function should have a global view of the expression and identify for which subexpression of the AND node the dominant set is required. By analyzing both the full expression and the specific AND node expression, the dominant set can be determined. The suggested implementation of the `getDominantSet` function would take two parameters: the entire expression and the specific AND node expression for which the dominant set is being found.

PR #144, opened 2024-10-12
done with subtract redundant constraint transformation

YeabsiraNigusse:
## Description

I have added a function called setDifference and  test functions for it. it is used for making a  subtract redundant constraint transformation over a given expression. This function assumes it receives a set1  and the set2 expression as parameters. It then removes redundant elements from set1 expression that already exist in the set2. This function also can be used for lCCSUBTRACT transformation with command set and current expression given as a parameter. The function uses isMember from general helpers functions.

## Motivation and Context

It is one of the main transformations applied to an expression, and it solves the problem of subtracting redundant constraints from the expression. 

It does not fix any existing issues but creates [this](https://github.com/iCog-Labs-Dev/metta-moses/issues/143) new issue.
## How Has This Been Tested?

i have added functions inside `general-helper-function-test file`. the file contain a unit test with different test cases and checks the function output with `assertEqual` built-in metta function.

No Environment needed. running the the test file would be enough.

## Types of changes

## Checklist:


reduct/boolean-reduct/redundant.metta
```diff
@@ -0,0 +1,27 @@
+
+; This function assumes it receives a dominant set and the current expression as parameters. 
+; It then removes redundant guard sets from the current expression that already exist in the dominant set. 
+; The dominant set should be the set of guard sets of the parent AND nodes for the current AND node.
+; This function also can be used for lCCSUBTRACT transformation with command set and current expression given as a parameter.
+; The function uses isMember from general helpers functions.
+
+
+(: setDifference (-> Expression Expression Expression))
```
Yagth, line 9: Dude. This function shouldn't be named `setDifference` it can confuse people in the long run. Can you rename it?

Yagth:
@samwondim , Yab claims that this function can be beneficial for you, can you take a look/

reduct/boolean-reduct/redundant.metta
YeabsiraNigusse: @Yagth this is what it says on the RTE function and what it does. its a function given for redundant transformation. should i call it `removeRedundant`.

Yagth: Best if you can do that, since it's actions corresponds to that transformation :)

Yagth:
@YeabsiraNigusse, so does the issue #143 still persists or it no longer needs to be addressed?

utilities/general-helpers.metta
```diff
@@ -47,3 +47,27 @@
...
+(= (setDifference $set1 $set2) 
+    (if (== $set1 ()) () ; Base case: return empty set when $set1 is empty
+        (let* (
+            ($head (car-atom $set1))  ; Extract the head of set1
+            ($tail (cdr-atom $set1))  ; Extract the tail of set1
+            ($newtail (setDifference $tail $set2))  ; Recursively process the tail
+        )
+            ; Check if the $head is in set2, if so, skip it
+
+            (if (and (isMember $head $set2) (not (== (get-metatype $head) Expression)) )
```
Yagth, line 65: Hey @YeabsiraNigusse, why do we need to make sure that the `$head` isn't an expression? Shouldn't the set difference work with any type of atom as an element in the set to be considered generic?

utilities/general-helpers.metta
```diff
@@ -47,3 +47,27 @@
...
+                $newtail 
+                ; Otherwise, add the head to the new tail, avoiding unnecessary parentheses
+                (if (and (== (car-atom $head) NOT) (isMember $head $set2) ) $newtail
```
Yagth, line 68: And can we not optimize the code by assigning the call to the `isMember` function to a variable?

Yagth (changes requested):

YeabsiraNigusse:
> @YeabsiraNigusse, so does the issue #143 still persists or it no longer needs to be addressed?

Yeah, I think we no longer need the issue since the transformation is now represented by a general helper function. For the `getDominant` function case, we could take another approach, such as updating the dominant set each time the expression changes, rather than recalculating it every time. This is more computationally efficient. However, if we can't do this due to the complexity of the function (RTE), we might need to reconsider it. We discussed this approach in the last meeting, so for now, we can consider the issue resolved.

utilities/general-helpers.metta
YeabsiraNigusse: @Yagth . Yes you are right! i was considering one RTE function case but yeah if it is a utility function it should be generic. i pushed another change. you can take a look.

utilities/general-helpers.metta
YeabsiraNigusse: @Yagth , i don't understand. what do you mean by that we can `optimize by assigning the call to the isMember function to a variable?`. could u please suggest me a way to do that.

Yagth: You can store the result of the `isMember` function to a variable inside the `let*` binding. This will save you the redundant computation of the function with the same parameters on this line.

Yagth (approved):

PR #145, opened 2024-10-14
implemented the cut-unnecessary-or transformation

Liya-F:
## Description
I have implemented the cut-unnecessary-or transformation. The POA for this transformation is any `OR` node. For this transformation to happen, the POA must have exactly one child. If so, the POA and its child will be removed, and the guardSet and children of the child are returned to the parent. In my implementation, I return the guardSet and grandChildren as a tuple each being a tuple. 

## Motivation and Context
This is one of the RTE transformations listed in the Holman paper. It is responsible for removing unnecessary OR nodes found in our expression. 

## How Has This Been Tested?
I have tested the function using different test cases and the `assertEqualToResult` method in meTTa. Currently, the function is unrelated to other functionalities in our codebase so it has not affected other parts. 

## Types of changes

## Checklist:

### DIFF
--- a/reduct/boolean-reduct/cut-unnecessary-or.metta
+++ b/reduct/boolean-reduct/cut-unnecessary-or.metta
@@ -15,10 +15,8 @@
         (
             (AND $exp)
             (NOT $exp)
-            (OR (if (== $restChildren ()) ((getGuardSetExp $child $child ()) $grandChildren) 
-            $exp))
+            (OR (if (== $restChildren ()) $grandChildren $exp))
         )
         )           
     )
 )
-
--- a/reduct/boolean-reduct/cut-unnecessary-or.metta
+++ b/reduct/boolean-reduct/cut-unnecessary-or.metta
@@ -1,0 +1,24 @@
+;a function which implements the Cut-Unnecessary-OR Transformation
+;POA: Any OR node
+;If the POA has only 1 child, orCut returns the child's guardSet and children to the POA's parent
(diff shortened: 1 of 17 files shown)


reduct/boolean-reduct/cut-unnecessary-or.metta
```diff
@@ -0,0 +1,24 @@
...
+            ($op (car-atom $exp))
+            ($children (cdr-atom $exp))
+            ($child (car-atom $children)) 
+            ($restChildren (cdr-atom $children))  
+            ($grandChildren (if (== (get-metatype $child) Symbol) () (cdr-atom $child)))  
+        )
+        (case $op
+        (
+            (AND $exp)
+            (NOT $exp)
```
Nahom32, line 17: @Liya-F  why did you specify both AND and NOT on your case statements?

Liya-F: > @Liya-F why did you specify both AND and NOT on your case statements?

Because the POA is an OR node, if the expression starts with AND or NOT, it is returned as it is. Should I only specify the condition for an OR node and put the rest inside $else?

Nahom32: @Liya-F it is fine It seemed to me as a redundant condition

Liya-F:
> Change the returned atom of the expression to be only the subexpression in stead of including the guardset too.

I have made the changes. It now returns the guardSet and Children together. Simply, it returns the grandChildren.

Yagth (approved):

PR #148, opened 2024-10-14
feat: enchance conacatTuple and isMember func

### LINKED ISSUE #134: Optimize the `isMember` function
**Detailed Description**

The `isMember` function used in the [rte-helpers.metta](https://github.com/iCog-Labs-Dev/metta-moses-reduction/blob/dev/reduct/boolean-reduct/rte-helpers.metta) file and defined in the [general-helpers.metta](https://github.com/iCog-Labs-Dev/metta-moses-reduction/blob/dev/utilities/general-helpers.metta) file can be optimized since the operation can be done in parallel utilizing the non deterministic nature of MeTTa. The current implementation is iterating over the expression to determine whether an atom exists in a given expression or not.

**Context**

Refactoring the function as such:
- reduces the overhead of using `car-atom` and `cdr-atom`.
- have better performance if done in parallel.

**Possible Implementation**

A possible implementation could be utilizing the builtin `intersection` function found in metta. (Documentation about this and similar other functions can be found [here](https://github.com/trueagi-io/hyperon-experimental/blob/main/lib/src/metta/runner/stdlib.metta)). One can try to find the intersection of the input atom and the rest of the atom to make sure the atom exists in there or not.

### LINKED ISSUE #135: Implement alternative `concatTuple` function
**Detailed Description**

The current implementation of `concatTuple` defined in [general-helpers.metta](https://github.com/iCog-Labs-Dev/metta-moses-reduction/blob/dev/utilities/general-helpers.metta) and used in several other places utilizes `car-atom` and `cdr-atom` to concatenate two tuples. We can however have a different implementation using the built in `union` function defined in MeTTa. (Documentation about this and other similar functions can be found [here](https://github.com/trueagi-io/hyperon-experimental/blob/main/lib/src/metta/runner/stdlib.metta)). 

The current `union` function in MeTTa version `0.1.11` and `0.1.12` preserves the order of the input atom even after the operation which is the behavior we want. However, this feature might be removed in later versions and the order might as well be disturbed. But better to use it while it's still here. And we will still have our previous implementation and use it in the upcoming MeTTa version when the `union` function's feature gets removed.

After the implementation of this function, our code base should also be refactored so that the new function is used instead of the old `concatTuple`.

**Context**

The implementation of this function should:
- yield a performance increase since the `concatTuple` function is used in different places.
- simplify our code.

Nardos-serkalem:
## Description

Implemented an alternative concatTuple using the built-in union function for better performance and code simplicity. Refactored the codebase to replace the old function with the new implementation.

This PR closes the following issues:
- Closes #134
- Closes #135

## Types of changes

## Checklist:

### DIFF
--- a/utilities/general-helpers.metta
+++ b/utilities/general-helpers.metta
@@ -25,16 +25,6 @@
 
 ; a helper function to the isConsistentExp function.
 ; a function which checks if an element is member of a tuple.
-(= (isMember $x $tuple)
-    (if (== $tuple ()) False
-        (let*
-            (
-                ($head (car-atom $tuple))
-                ($tail (cdr-atom $tuple))
-            )
-        (
-            if (== $x $head) True (isMember $x $tail)
-        )
-)
-)
+(= (isMember $x $list)
+    (not (== (collapse (intersection (superpose ($x)) (superpose $list))) ()))
 )
--- a/utilities/general-helpers.metta
+++ b/utilities/general-helpers.metta
@@ -4,17 +4,8 @@
 (= (~= $a $b) (not (== $a $b)))
 
 ;Function to cocatinate two tuples (A B) (C D) ==> (A B C D)
-;(: cocatTuple (-> Expression Expression Expression))
- (= (concatTuple $xs $ys)
-   (if (== $xs ())
-       $ys
-       (let* (($head (car-atom $xs)) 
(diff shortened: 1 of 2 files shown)


Yagth (approved):

Yagth:
@Nardos-serkalem , can you open the same PR to the dev branch instead?

PR #151, opened 2024-10-17
optimize-isConsistenetExp

### LINKED ISSUE #133: Optimize the `isConsistentExp` function
**Detailed Description**

The `isConsistentExp` function used in the [delete-inconsistent-handle.metta](https://github.com/iCog-Labs-Dev/metta-moses-reduction/blob/dev/reduct/boolean-reduct/delete-inconsistent-handle.metta) file and defined in the [rte-helpers.metta](https://github.com/iCog-Labs-Dev/metta-moses-reduction/blob/dev/reduct/boolean-reduct/rte-helpers.metta) file can be optimized since the operation can be done more optimally by utilizing the non deterministic nature of MeTTa. The current implementation is iterating over the expression to determine whether an atom's negation exists in the expression or not one at a time.

**Context**

Refactoring the function as such:
- reduces the overhead of using `car-atom` and `cdr-atom` for iteration over the atom one at a time.
- have better performance if done in parallel.
- the code might even be more elegant since the complexity decreases.

**Possible Implementation**

A possible implementation could be defining a non deterministic function that takes two atoms at a time. One can then duplicate the expression in question and pass it to this function after superposing each. This makes sure that the function gets every possible combination of the atoms in the expression. i.e., `((superpose (A B) (superpose (A B))` will return `[(A A), (A B), (B A), (B B)]` which is every combination of the expression that there is.

The defined function can then compare if the two atoms are negations of one another to decide whether the original expression was consistent or not.

tarikuamisganaw:
## Description

I have optimized the isConsistentExp function based on the suggested implementation in the issue. The optimization leverages the non-deterministic nature of MeTTa, reducing overhead and improving performance. I defined a non-deterministic function that processes two atoms at a time. The expression in question is duplicated and passed to this function after applying superpos, ensuring that all possible combinations of the atoms in the expression are checked.

This PR closes the following issue:
- Closes #133 

## Motivation and Context
Improved performance by eliminating the need to manually iterate over atoms with car-atom and cdr-atom.
The code is more elegant and less complex due to reduced overhead and parallel execution.

## How Has This Been Tested?

I tested the changes using the existing tests in the helper-function-test.metta file to ensure that all test cases passed successfully.

## Types of changes

Optimization of the isConsistentExp function, which is used in the delete-inconsistent-handle.metta file and defined in the rte-helpers.metta file.

## Checklist:

[X] My code follows the project's code style.
[X] All new and existing tests passed.

### DIFF
--- a/reduct/boolean-reduct/rte-helpers.metta
+++ b/reduct/boolean-reduct/rte-helpers.metta
@@ -35,21 +35,30 @@
     )
 )
 
-;a function to check whether an n-ary expression is consistent or not.
+;function to check consistency of  an expression
 (= (isConsistentExp $exp)
-(let $guardSetTuple (getGuardSetExp $exp $exp ()) 
-(if (== $guardSetTuple ()) True
-        (let*
-            (
-                ($head (car-atom $guardSetTuple))
-                ($tail (cdr-atom $guardSetTuple))
-            )
-        (if (isMember (Not $head) $tail) False
-            (isConsistentExp $tail)
-        ))
-)
-)   
-)
+   (let $guardSetTuple (getGuardSetExp $exp $exp ())
+      (if (== $guardSetTuple ()) True
+          (check-negations (collapse ((superpose $guardSetTuple) (superpose $guardSetTuple))))
+             )))
+
+;function to check negation pairs to check each pair in a list of a tuple
+(= (negation-pair $a $b)
+   (or (== $a (NOT $b)) (== $b (NOT $a))))
+
+;function  to recurse through the list of the tuple to check each pair
+(= (check-negations $pairList)
+   (if (== $pairList ()) 
+       True
+       (let* (
+        ($pair (car-atom $pairList))     
+        ($first (car-atom $pair))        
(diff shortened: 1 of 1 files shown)


reduct/boolean-reduct/rte-helpers.metta
```diff
@@ -35,21 +35,30 @@
...
+      (if (== $guardSetTuple ()) True
+          (check-negations (collapse ((superpose $guardSetTuple) (superpose $guardSetTuple))))
+             )))
+
+;function to check negation pairs to check each pair in a list of a tuple
+(= (negation-pair $a $b)
+   (or (== $a (NOT $b)) (== $b (NOT $a))))
+
+;function  to recurse through the list of the tuple to check each pair
+(= (check-negations $pairList)
```
Yagth, line 50: @tarikuamisganaw, I believe there is still a way to remove the recursion used here. Can you update that?

Yagth (changes requested):

Yagth:
@tarikuamisganaw just merged another PR aimed at solving the same issue. This PR is no longer needed.

PR #159, opened 2024-10-24
Fixes #138

### LINKED ISSUE #138: Refactor `getGsetAndChildren` function
**Detailed Description**

The current implementation of `getGsetAndChildren` function defined in [rte-helpers.metta](https://github.com/iCog-Labs-Dev/metta-moses-reduction/blob/dev/reduct/boolean-reduct/rte-helpers.metta) and used in different places utilizes another function `GsetandChildren` function which in turn uses `car-atom` and `cdr-atom` to iterate over a given expression and distinguish between `guardSets` and `children` of that expression. The function assumes that both `AND` and `OR` nodes have `guradSets` of similar interpretation. It then returns a tuple of the `guardSets` and children found in the expression. i.e., for an input expression `(AND A B (OR A B) C)` it will return `((A B C) ((OR A B))`. The function returns the same result even if the expression started with an `OR` node. 

We can, however, have an optimized function that does the same thing with improved performance.

**Context**

The implementation of this function should:
- yield a performance increase.
- simplify our code.
- reduce code redundancy by removing functions that does similar but slightly different things.

**Possible Implementation**
A possible implementation could be utilizing the refactored `getGuardSetExp` function in issue #132 and another function for getting the children in an expression. The latter function could be implemented in a way similar to the `getGuardSetExp` with a slight variation or it could be implemented using a different logic if a better one is found. (*Decision left to the person working on this issue.*)

Amicahh:
- optimized function that does the same thing with improved performance.
- The implementation of this function does:
  - yield a performance increase.
  - simplify the code.
  - reduce code redundancy by removing functions that does similar but slightly different things.

### DIFF
--- a/reduct/boolean-reduct/rte-helpers.metta
+++ b/reduct/boolean-reduct/rte-helpers.metta
@@ -93,7 +93,66 @@
 )
 
 (= (getGsetAndChildren $expr) (GsetandChildren $expr () ()))
+; Function to get the guardSet and children from an expression.
+(: getGsetAndChildren (-> Expression Expression))
+(= (getGsetAndChildren $expr)
+     (getGsetAndChildrenHelper $expr () ()))  ; Initial call with empty tuples for guardSets and children.
 
+; Helper function to recursively find guardSets and children.
+(: getGsetAndChildrenHelper (-> Expression Expression Expression Expression))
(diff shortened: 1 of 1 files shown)


samwondim:
removing unwanted pr

PR #161, opened 2024-10-31
Chore/improved functions

Yagth:
## Description

I have added comments for each transformation function about what it is doing. The documentation is mostly taken from the holman paper and modified to match the behavior of the functions. Also, other minor updates like formatting, function naming and some function locations have been made. The functions that were defined in the `rte-helpers.metta` file but are only used in a single file have been moved to that file.

I have also added two more test cases that were edge cases for the `promote-common-constraint` and `zero-constraint-subsumption` transformation rules. They are commented out for now but a fix is under way.

I have also add few helper functions that will be useful for rte implementation.

## Motivation and Context

The main motivation of this context is to keep the code consistency and improved code readability.

## How Has This Been Tested?

I run all the test cases locally and everything is working fine after the update.

## Checklist:

### DIFF
--- a/utilities/general-helpers.metta
+++ b/utilities/general-helpers.metta
@@ -51,3 +51,9 @@
         )
     )
 )
+
+(= (replace $old $new $current) (if (== $old $current) $new $current))
+
+(= (findAndReplace $old $new $list)
+    (collapse (replace $old $new (superpose $list)))
+)
--- a/utilities/general-helpers.metta
+++ b/utilities/general-helpers.metta
@@ -57,3 +57,7 @@
 (= (findAndReplace $old $new $list)
     (collapse (replace $old $new (superpose $list)))
 )
+
+
+(: until (-> (-> $a Bool) (-> $a $a) $a $a))
+(= (until $predicate $function $x) (if ($predicate $x) $x (let $x' ($function $x) (until $predicate $function $x'))))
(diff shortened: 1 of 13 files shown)


utilities/general-helpers.metta
```diff
@@ -57,3 +57,7 @@
 (= (findAndReplace $old $new $list)
     (collapse (replace $old $new (superpose $list)))
 )
+
+
+(: until (-> (-> $a Bool) (-> $a $a) $a $a))
+(= (until $predicate $function $x) (if ($predicate $x) $x (let $x' ($function $x) (until $predicate $function $x'))))
```
Nahom32, line 63: @Yagth is this implementation concerned with executing a function until a truth value is met?

utilities/tests/general-helper-functions-test.metta
```diff
@@ -38,3 +38,59 @@
...
+;; These are functions used for testing until function.
+
+;; It returns True only if the input is zero.
+(: isZero (-> Number Bool))
+(= (isZero $x) (== $x 0))
+
+;; It will return True if greater than 10.
+(: greaterThanTen (-> Number Bool))
+(= (greaterThanTen $x) (> $x 10))
+
```
Nahom32, line 53: why do you need this function? I mean it is a trivial check and very specific

Nahom32 (changes requested):

utilities/general-helpers.metta
Yagth: Yep. I was planning to use it like a while loop for some cases if possible

utilities/tests/general-helper-functions-test.metta
Yagth: Yeah it is. I only need it to for testing the `until` function, hence the reason I put it in the test file.

Nahom32 (approved):

PR #163, opened 2024-11-28
Create 1-constraint-complement-subtraction.metta

K1rubel:
The function oneCcSub applies the 1-constraint-complement-subtraction transformation to the sub-expression given. A literal in the guard set of a non-root AND node is removed from the node if its negated form is present in the guard set of a commanding and node, COM.

### DIFF
--- a/reduct/boolean-reduct/1-constraint-complement-subtraction.metta
+++ b/reduct/boolean-reduct/1-constraint-complement-subtraction.metta
@@ -1,30 +1,32 @@
 ;; Name:                oneCcSub
-;; A function to remove an element from the point of application (POA) if it's negated form is 
-;;                      contained in the guard set of a commanding AND node (COM).
+;; A function to remove an element from the point of application (POA) if its negated form is contained in the
+;;                       guard set of a commanding AND node (COM).
 ;; Parameters:
-;;                      1. The guard set of the commanding terminal AND node (COM), $comSet.
-;;                      2. The point of application, $poa, a non-root AND node
-;;                      3. The parent expression,$parent, possibly containing the point of application 
-;;                      and a commanding AND node.
-;; Goal:                To remove a constraint that is not needed because it is commanded by 
-;;                      its negated form.
+;;                      1. $parent:- the parent expression, possibly containing the point of application and a commanding AND node.
+;;                      2. $poa:- the point of application, a non-root AND node.
+;;                      3. $com:- the guard set of the commanding terminal AND node (COM).
+;;                      
+;; Goal:                To remove a constraint that is not needed because it is commanded by its negated form.
+;;                      
 ;; POA:                 Any non-root AND node.
+;;
 ;; Preconditions:       The POA’s guard set contains constraint X and the POA is commanded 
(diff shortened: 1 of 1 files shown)


reduct/boolean-reduct/1-constraint-complement-subtraction.metta
```diff
@@ -0,0 +1,57 @@
...
+;; POA:                 Any non-root AND node.
+;; Preconditions:       The POA’s guard set contains constraint X and the POA is commanded 
+;;                      by a terminal AND node COM whose guard set consists of the constraint 
+;;                      which is the opposite of X.
+;; Action:              Subtract the constraint X from the guard set of the POA.
+;;
+;; Example:             POA --> (AND B C), COM guard set ((NOT B) A) -- a negated literal of the POA is                        found in the COM guard set so the POA is commanded by this set and the common
+;;                      element is subtracted from the POA that member, B in this case. After the 
+;;                      transformation, the POA is now (AND C).
+(= (oneCcSub $comSet $poa $subExpr)
```
Yagth, line 15: It's not clear what the `$subExpr` is so can you add a comment or a corresponding example in addition to the POA in the above comment?

reduct/boolean-reduct/1-constraint-complement-subtraction.metta
```diff
@@ -0,0 +1,57 @@
...
+(= (expressionEquality $a $b)
+    (let $diff (collapse (subtraction (superpose $a) (superpose $b)))
+        (if (== $diff ())
+            True
+            False
+        )))
+
+;; test cases
+
+! (oneCcSub ((NOT B) C) (AND B C) (OR (AND D) (AND B C))) ;(OR (AND D) (AND C))
```
Yagth, line 53: I think you forgot to extract these into their own separate test files like all the other transformation functions? This file shouldn't contain any executable

reduct/boolean-reduct/1-constraint-complement-subtraction.metta
Yagth: May be you can rename it to `parent` if that's the case. But one last remark I have here is that the other transformation functions have parameter order of `parent`, `poa` and `comSet`. It's best to keep this same pattern for the sake of consistency.

K1rubel: I have edited the file accordingly.

reduct/boolean-reduct/1-constraint-complement-subtraction.metta
K1rubel: Removed the test cases!
Need I do another pull request for the tests?

Yagth: No you can commit the changes and push it to your branch. I think the PR will be updated automatically then

reduct/boolean-reduct/1-constraint-complement-subtraction.metta
Yagth: Cool, can you reorder the parameters to match the order `parent`, `poa` and `com`?

PR #172, opened 2024-12-06
feat: implemented scoring

eyuuab:
Implement scoring for boolean expression based on Accuracy and Complexity


scoring/fitness.metta
```diff
@@ -0,0 +1,83 @@
...
+! (import! &self EDA:scoring:general-helpers)
+! (bind! &space (new-space))
+;! (import! &test testData)
+
+
+; measure the accuracy of the exprssion based on the given data 
+(= (fitness $expr $data)
+    (if (== $data ())
+        0
+        (let*(
```
Yagth, line 11: Can you add space after `*` and `(` here?

Yagth: And you could have just used `let` instead of `let*` right?

scoring/fitness.metta
```diff
@@ -0,0 +1,83 @@
...
+(= (penalizedFitness $expr $data $lambda)
+    (let* (
+        ($accuracy (fitness $expr $data)) 
+        ($complexity (counter $expr))    
+    )
+        (- $accuracy (* $lambda $complexity))  
+    )
+)
+
+;; Test Cases 1: Simple AND Expression
```
Yagth, line 44: Can you move these test cases to their own files? Or does that create issues?

scoring/fitness.metta
```diff
@@ -0,0 +1,83 @@
...
+            ($replacedExpr (replaceWithTruth $expr &space (collapse (match &space $x $x))))
+            ($evaluateuatedResult (evaluate $replacedExpr))
+            ($remove (collapse (removeFromSpace (superpose $inputs) &space)))
+        )
+        (if (== $evaluateuatedResult $target) 1 0)
+    )
+)
+
+
+
```
Yagth, line 34: Would be cool to have a small comment on this function :)

scoring/general-helpers.metta
```diff
@@ -0,0 +1,112 @@
+
+
+(= (addToSpace $row $space) (add-atom $space $row))
```
Yagth, line 3: Why did you need this function? Why not use `add-atom` it self instead of this new functions?

scoring/general-helpers.metta
```diff
@@ -0,0 +1,112 @@
...
+
+(= (removeFromSpace $row $space) (remove-atom $space $row))
```
Yagth, line 5: Same comment here. I still think this can be done using only the `remove-atom` function. Don't you think?

scoring/general-helpers.metta
```diff
@@ -0,0 +1,112 @@
...
+(= (foldr $f $i $xs)
+   (if (== $xs ())
+       $i
+       (let* (($h (car-atom $xs))
+              ($t (cdr-atom $xs))
+              ($ft (foldr $f $i $t)))
+         ($f $h $ft))))
+
+; replace expersion with Truth value from the space
+(= (replaceWithTruth $expr $space $spaceContent)
```
Yagth, line 17: I think this function is specific to the fitness function temporarily so it's best to move it there

Yagth: In fact. Except the `foldr` function, your newly added functions in this file should be moved to the `fitness.metta` file. If the functions defined here aren't being used by any other file than the `fitness.metta` file, its best to keep them in that file. For more details about this and other conventions we came up with you can refer the [Contributing File](https://github.com/iCog-Labs-Dev/metta-moses/blob/main/.github/CONTRIBUTING.md#code-conventions)

Yagth (changes requested):
Looking good. Just few lose ends to tie and then we're good for merging.

Yagth (approved):

PR #174, opened 2025-01-09
added recombination and mutation operators + some general expression …

K1rubel:
## Description
The recombination operator takes two parent genotypes and recombines genotype material from one parent to the other using a binary mask of the same size as the genotype's length. A mask value of 0 represents no change and 1 means the genotype at that position will replace and at the same time be replaced by genotype information from the other parent and vice versa. 
The mutation operator takes a genotype and introduces mutation to a specific gene of a non-terminal selected randomly from a list of non-terminals. 

## Motivation and Context
These two operators represent means of exploring solution space in the boolean expression creation procedure used by Structured Grammatical Evolution (SGE) hoping to produce better-performing offspring.
Recombination: takes genetic material from two parents and creates two crossbreeds. 
Mutation: makes changes to a specific gene in the genotype of a non-terminal. 

## How Has This Been Tested?
Recombination and mutation operations have been tested locally by producing new crossbreeds and changing a gene in the genotype of a non-terminal. 

## Types of changes

## Checklist:


K1rubel:
The assertEqual for **_mutationGeneCount_** is intolerably slow since every function (main + helpers) employed here has been written to be fully recursive to avoid the surprise factor in a _non-deterministic_ approach using superpose and such ..

K1rubel:
The _assertEqual_ is still slow but less intolerable now 😁

algorithms/sge.metta
```diff
@@ -641,6 +642,114 @@
...
+(= (findNt $grammarSpace)
+    (let* 
+        (
+            ($ntSet (getNTs $grammarSpace))
+            ($len (len $ntSet))
+        )
+            (√nt $len $ntSet $grammarSpace)))
+
+;; √nt makes sure the picked NT has expansion option ≥ 2
+(= (√nt $len $ntSet $grammarSpace)
```
Yagth, line 723: I love the simplicity of the name but won't this cause issue if someone wants to call the function? I don't think the "tick" is easily accessible on our keyboards :)

utilities/expression-helpers.metta
```diff
@@ -113,3 +113,38 @@
...
+                ($first (car-atom $expr))
+                ($tail (cdr-atom $expr))
+                ($carry (++ $acc $first))
+            )
+            (___f $tail $carry)
+            )))
+
+;; mutatedGeneCount counts the number of mutated genes
+
+(= (mutatedGeneCount $original $grammarSpace)
```
Yagth, line 136: I don't think mutating with in this function should be the expected behavior of this function. Why not accept the mutated gene too since the function's purpose is just to count the number of mutated genes. right?

utilities/general-helpers.metta
```diff
@@ -207,3 +207,153 @@
...
+                ($ft (cdr-atom $first))
+                ($s (car-atom $second))
+                ($st (cdr-atom $second))
+                ($r  (== $f $s))
+                ($c (compareElements $ft $st))
+            )
+                    (cons-atom $r $c)
+                )))
+
+;; ++n counts occurence of $item in $expr
```
Yagth, line 345: We have our own `count` function already implemented, why not use that?

utilities/general-helpers.metta
```diff
@@ -207,3 +207,153 @@
...
+
+(= (selectByIndex $expr $index)
+    (if (== $expr ())
+        -1
+        (if (== $index 0)
+            (car-atom $expr)
+            (selectByIndex (cdr-atom $expr) (- $index 1))
+        )))
+            
+(= (getNTs $grammarSpace)
```
Yagth, line 307: This function could be done in oneliner in an elegant way if you can just use `match`

utilities/general-helpers.metta
```diff
@@ -207,3 +207,153 @@
...
+                (let $c (replaceByIndex $tail (- $index 1) $new)
+                    (cons-atom $first $c)
+                )))))
+
+;; Select atom by index
+;; description:-        0 indexed selection of atom in a tuple
+;;                      Returns -1 if index is out of range
+;;                      If $index 0 -- firs element is returned (car-atom $expr)
+;;                      Otherwise calls selectByIndex recusively with the rest of $expr, (cdr-atom $expr) and (- $index 1)
+;; Example
```
Yagth, line 294: The example and description isn't necessary for this function, it is pretty much self explanatory. Just add a comment like: "Select atom by index and return -1 if index is out of range".

utilities/general-helpers.metta
```diff
@@ -207,3 +207,153 @@
...
+;;                      ! (selectByIndex (0 1 2 3 4) 5) --> -1 , index out of range
+;;                      ! (selectByIndex (0 1 2 3 4) 0) --> 0
+;;                      ! (selectByIndex ((0 1) (2 3 4) (1 0 1)) 1) --> (2 3 4)
+
+(= (selectByIndex $expr $index)
+    (if (== $expr ())
+        -1
```
Yagth, line 301: What if we return an error here? Like `(Error index out of range or something)`. I said that because this is the expected behavior with other programming languages and since this function may be used later on it is best to stick with standard implementations :)

utilities/general-helpers.metta
```diff
@@ -207,3 +207,153 @@
...
+;;          if $expression is a multi-element expression
+;;          check if index == 0, if so append the new $val to $acc and append the this to the rest of expression, $tail
+;;          if $index not 0, call the function with (- $index 1) and all the previous elements appended to $acc
+;; Examples 
+;;         ! (replaceByIndex (0 1 2 3 4) 0 7) -->             (7 1 2 3 4)
+;;         ! (replaceByIndex (0 1 2 3 4) 2 7) -->             (0 1 7 3 4)
+;;         ! (replaceByIndex ((0) (0 1) (2 3 4)) 1 (2 1)) --> ((0) (2 1) (2 3 4))
+
+(: replaceByIndex (-> Expression Number $t Expression))
+(= (replaceByIndex $expr $index $new)
```
Yagth, line 272: Again here, the behavior of `replaceByIndex` should be as expected, it shouldn't try to add a new element if there is no element at that index. I am saying this because this function can also be used in other contexts and it's best to stick with the standard behavior.

utilities/general-helpers.metta
```diff
@@ -207,3 +207,153 @@
...
+                (if (isLiteral $new)
+                    (cons-atom $first ($new))
+                    (cons-atom $first $new)
+                )
+                (let $carry (++ $tail $new)
+                    (cons-atom $first $carry)
+                )))))
+
+;; replaceByIndex replaces an atom at a specific index position with $val -- 0 (indexed counting is implied)
+;; How it works
```
Yagth, line 256: Again we don't need a description and example for this function too as it is a common function. Short comment should suffice.

utilities/general-helpers.metta
```diff
@@ -207,3 +207,153 @@
           )
           (cons-atom ($x $y) $rest)))
 )
+
+;; Finds length of a tuple
+
+(= (len $expr)
+    (if (== $expr ())
+        0
+        (+ 1 (len (cdr-atom $expr)))
```
Yagth, line 216: Just curious about this one, will it have effect on it's performance if you use the `let` binding here instead of calling the recursion like you did?

utilities/expression-helpers.metta
```diff
@@ -113,3 +113,38 @@
...
+        (
+            ($mutated (mutate $original $grammarSpace))
+            ; ($mutated ((0) (1 0 1) (1 2 3)))
+            ($o_ (___f $original ()))
+            ($m_ (___f $mutated ()))
+            ($bool (compareElements $o_ $m_))
+            ; (() (println! $bool))
+            ; ($n (++n $bool False 0))
+        )
+        (++n $bool False)
```
Yagth, line 147: You can use the built in count instead of this

utilities/expression-helpers.metta
```diff
@@ -113,3 +113,38 @@
...
+(= (mutatedGeneCount $original $grammarSpace)
+    (let*
+        (
+            ($mutated (mutate $original $grammarSpace))
+            ; ($mutated ((0) (1 0 1) (1 2 3)))
+            ($o_ (___f $original ()))
+            ($m_ (___f $mutated ()))
+            ($bool (compareElements $o_ $m_))
+            ; (() (println! $bool))
+            ; ($n (++n $bool False 0))
```
Yagth, line 145: Best if you remove this line. (It's called a zombie code. It can ruin our day someday in the future 😂 )

algorithms/sge.metta
```diff
@@ -641,6 +642,114 @@
...
+;;                          ! (recombine ((0) (1 0))((1) (1 1)) (1 0)) --> (((1) (1 0)) ((0) (1 1)))
+;;                          ! (recombine ((0) (1 0) (1 2 0)) ((1) (1 1) (1 1 2)) (0 1 1)) --> (((0) (1 1) (1 1 2)) ((1) (1 0) (1 2 0)))
+;;                          ! (recombine ((0) (1 0) (1 2 0) (1 1 0 0)) ((1) (1 1) (1 1 2) (0 0 1 1)) (0 1 1 0)) --> (((0) (1 1) (1 1 2) (1 1 0 0)) ((1) (1 0) (1 2 0) (0 0 1 1)))
+
+(= (recombine $parent1 $parent2 $mask)
+    (if (== $mask ())
+        ($parent1 $parent2)
+        (let*
+            (
+                ($gene1 (car-atom $parent1))
```
Yagth, line 665: Some of the following codes will be shorter if you can use the `decons` function in the `general-helpers` file

algorithms/sge.metta
Yagth: And I think I forgot the purpose of this function but we can talk over it when your at the office.

Yagth (changes requested):
Finish the changes I request and it will be good to go.

algorithms/sge.metta
K1rubel: Since the mutation agent was relaxed to allow for redundant mutations, there is no need for those functions, the sole purpose of them was to enforce mutation at all times. 
I understand the symbols are not ideal for anyone to use. 🙏🏾

utilities/expression-helpers.metta
```diff
@@ -113,3 +113,38 @@
     )
   )
 )
+
+;; assertEqual for mutation operator
+;; ___f flattens a tuple of tuples keeping the order
+;; Examples
+;;          ! (___f ((0 1) (1 2 1 0))) --> (0 1 1 2 1 0)
+
+(= (___f $expr $acc)
```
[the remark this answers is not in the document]
K1rubel: used `(foldr ++ () $tuple) `--  looks good  🙏🏾🙏🏾

algorithms/sge.metta
```diff
@@ -641,6 +642,114 @@
...
+;; Note:-       calls itself recursively till it finds a value within [0, cN) that is different from the old value at $index
+;;              This last point is subject to furthure review and modification based on findings about how redundant mutation in SGE should be handled
+;;              This implementation makes sure the new genotype is different from the parent genotype at all times
+
+(= (mutateGene $genes $index $cN)
+    (let* (
+        ($old (selectByIndex $genes $index))
+        ($new ((py-dot rnd randint) $cN))
+        )
+        (if (== $new $old)
```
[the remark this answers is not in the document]
K1rubel: I have edited the mutation in a way that allows generation of identical genotype as the starting genotype pure based on chance.

utilities/expression-helpers.metta
K1rubel: removed the _mutation_ agent from call within the _mutatedGeneCount_ function. Just used mutated genotype tuple as a parameter in call in the _assertEqual_ function.

utilities/general-helpers.metta
K1rubel: didn't know we had the `count` function. I have used it now. 🙏🏾

utilities/general-helpers.metta
K1rubel: `match` produces results in unpredictable order but the approach used here produces consistently ordered output, so far!

utilities/general-helpers.metta
K1rubel: 👍🏾

utilities/general-helpers.metta
K1rubel: It just returns the original expression if the index is out of range. I have tested again and works as expected.

utilities/general-helpers.metta
K1rubel: 👍🏾

algorithms/sge.metta
K1rubel: Didn't know we had that! looks cool! 🙏🏾

utilities/general-helpers.metta
Yagth: Oww I see why you did that.

utilities/general-helpers.metta
Yagth: Yeah my bad. I overlooked that.

utilities/expression-helpers.metta
```diff
@@ -113,3 +113,20 @@
     )
   )
 )
+
+;; mutatedGeneCount counts the number of mutated genes and 
+;;        return true only if the number of changed genes is either 0 or 1
+;;        Flattens the two gene lists, original and mutated, so that changes can be compared element by element
+
+(: mutatedGeneCount (-> Expression Expression Bool))
+(= (mutatedGeneCount $original $mutated)
```
Yagth, line 122: mutatedGeneCount now doesn't reflect the behavior of the function. Best if you can rename it. And this function is specific for the testing only right? So it's best to keep it just above the test cases you wrote for now. In the near future, when we can put the test cases into their own separate file, this function will be in the test file it self. :)

PR #203, opened 2025-04-01
Exemplar and metapopulation data structures

K1rubel:
Data structure definitions for exemplar and metapopulation. 
Added function getting scores from an exemplar, comparing exemplars, inserting an exemplar into metapopulation (represented here as an ordered set).

## Types of changes

## Checklist:

### DIFF
--- a/utilities/ordered-set.metta
+++ b/utilities/ordered-set.metta
@@ -1,0 +1,88 @@
+;; data structure for exemplars -- aka scored trees -- Xmplr
+;; deme ID
+(: demeID Type)
+(: mkDemeID (-> Number demeID))
+
+;; composite_score
+(: cscore Type)
+(: getCscore (-> tree cscore))
+        
+;; composite score components
+(: scoreT Type)
+(: cpxyT Type)
+
+(: mkScoreT (-> Number scoreT))
+(: mkCpxy (-> Number cpxyT))
+
+;; composite score is composed of a raw score, comlexity score, a complexity penalty and a uniformity penalty score
+(: mkCscore (-> scoreT cpxyT scoreT scoreT cscore))
+
+;; behavioral_score
+(: bscore Type)
+(: mkBscore (-> Number bscore))
+
+(: Xmplr Type)
+
+(: mkXmplr (-> tree demeID cscore bscore Xmplr))
+
+;; XXX - SUGGESTION -- for the exemplars, I have tried to follow the type definitions as they are defined in C++ MOSES
+;; although now I don't exactly see the advantage using specific types for numerical values such as bscore, score ..
+;; I say it is better better to uses these as Numbers without wrapper type definitions
+
+;; data structures for the metapopulation
+;; keeping the same naming convention as used in the definitio of the ordered multiset
+;; Ordered set data structure -- OS
+
+(: OS (-> $a Type))
+(: mkOS (-> $a (OS $a)))
+
+;; the metapopulation -- MP as an ordered set of exemplars (OS Xmplr)
+(: MP (OS Xmplr))
+
+(: mkMP (-> Xmplr (OS Xmplr) (OS Xmplr)))
+(: NilXmplr (OS Xmplr))
+
+;; compare two exemplars in a metapopulation using penalized score using complexity score to separate ties
+;; G -- if first exemplar greater than the second in terms of penalized score or if this is the same lower complexity score 
+;; E -- exemplars equal in both of the scores
+;; L -- either lower in penalized score or higher complexity with equal penalized scores
+
+(MP.XmplrCompare (-> Xmplr Xmplr Atom))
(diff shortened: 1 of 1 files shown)


utilities/ordered-set.metta
```diff
@@ -0,0 +1,88 @@
...
+
+(: mkMP (-> Xmplr (OS Xmplr) (OS Xmplr)))
+(: NilXmplr (OS Xmplr))
+
+;; compare two exemplars in a metapopulation using penalized score using complexity score to separate ties
+;; G -- if first exemplar greater than the second in terms of penalized score or if this is the same lower complexity score 
+;; E -- exemplars equal in both of the scores
+;; L -- either lower in penalized score or higher complexity with equal penalized scores
+
+(MP.XmplrCompare (-> Xmplr Xmplr Atom))
```
Yagth, line 50: Where is this comparing function located?

utilities/ordered-set.metta
```diff
@@ -0,0 +1,88 @@
...
+                (case ((> $p1 $p2) (== $p1 $p2) (< $cs1 $cs2) (== $cs1 $cs2))
+                    
+                    (((True $t $u $v) G)
+                     ((False True True $v) G)
+                     ((False True False True) E)
+                     ((False True False False) L)
+                     ((False False $u $v) L) ))))
+
+;; get penalized score of an exemplar
+(: Xmplr.getPnzdScr (-> Xmplr ScoreT))
```
Yagth, line 66: This kind of function abrivations are hard to remember. It's best if you use the whole name instead

utilities/ordered-set.metta
```diff
@@ -0,0 +1,88 @@
...
+    (let ($tree $demeID ($s $cs $cp $up) $bscore) $xmplr  (- (- $s $cp) $up)))
+
+;; get expression of probability values 
+(: MP.getProbs (-> (OS Xmplr) (Exp ScoreT)))
+(= (MP.getProbs $mp)
+    (collapse 
+        (let $xmplr (superpose $mp) (Xmplr.getPnzdScr $xmplr))))
+
+;; insert an exemplar (Xmplr) in to the metapopulaion
+(: MP.insert (-> Xmplr (OS Xmplr) (OS Xmplr)))
```
Yagth, line 77: Best if the insert function is actually part of the ordered set and generic. The insert function for example can take a comparing function to do it's logic properly. If you do that, then this kind of function definitions won't be needed in the future.

utilities/ordered-set.metta
```diff
@@ -0,0 +1,88 @@
...
+                     ((False True False False) L)
+                     ((False False $u $v) L) ))))
+
+;; get penalized score of an exemplar
+(: Xmplr.getPnzdScr (-> Xmplr ScoreT))
+(= (Xmplr.getPnzdScr $xmplr)
+    (let ($tree $demeID ($s $cs $cp $up) $bscore) $xmplr  (- (- $s $cp) $up)))
+
+;; get expression of probability values 
+(: MP.getProbs (-> (OS Xmplr) (Exp ScoreT)))
```
Yagth, line 71: Do you think this function will be used else where?

utilities/ordered-set.metta
```diff
@@ -0,0 +1,88 @@
...
+
+;; behavioral_score
+(: bscore Type)
+(: mkBscore (-> Number bscore))
+
+(: Xmplr Type)
+
+(: mkXmplr (-> tree demeID cscore bscore Xmplr))
+
+;; XXX - SUGGESTION -- for the exemplars, I have tried to follow the type definitions as they are defined in C++ MOSES
```
Yagth, line 28: The type definitions were actually needed to help us deferentiate what type of each number is. This can help prevent accidental variable passing with in our logic as we will get additional help from the type checker.

utilities/ordered-set.metta
```diff
@@ -0,0 +1,88 @@
...
+;; although now I don't exactly see the advantage using specific types for numerical values such as bscore, score ..
+;; I say it is better better to uses these as Numbers without wrapper type definitions
+
+;; data structures for the metapopulation
+;; keeping the same naming convention as used in the definitio of the ordered multiset
+;; Ordered set data structure -- OS
+
+(: OS (-> $a Type))
+(: mkOS (-> $a (OS $a)))
```
Yagth, line 37: Also I think the type definition should be a recursive one like list. Refer eyobed's multiset definition in dev branch to see what I mean.

Yagth: Also, this file should contain information only about the ordered set since the file name is `ordered-set.metta`. The other logic should be in a different file like `mosesTypes.metta` for example.

Yagth (changes requested):
Few changes you need to make before merging it. Also don't forget to add simple test cases for the functions you added.

PR #211, opened 2025-04-01
fixed the issue on concatTuple function

### LINKED ISSUE #204: concatTuple function stopped working.
**Description**

The `concatTuple` expression used in the previous implementations now stopped working. Previously it was implemented using non deterministic function `union` to increase performance. But now the `union` function behavior has changed and the MeTTa interpreter's speed has increase so this function can be reimplemented using recursion.

After the fix, the test cases corresponding to the commented out in the `utilities/tests/general-helper-functions-tests.metta`  should pass.

**Expected Behavior**

The function should have merged two expressions together preserving their order.

**Actual Behavior**

the function changes order of the expressions being merged. The unit tests commented out in the `utilities/tests/general-helper-functions-tests.metta` that correspond to this function are failing. 

**Possible Fix**

**Your Environment**

* Version used:
* MeTTa version (e.g. 0.1.11): 0.2.3

blackhammer116:
## Description
The concatTuple wasn't working in the previous non-deterministic implementation with the new version. The changed implementation now utilizes the ```union-atom``` to concat the two tuples given while also preserving their order. However this implementation only works if the script is run inside the virtual environment using just ```metta``` instead of ```metta-run``` 

## How Has This Been Tested?
By running the commented test cases inside ```utilities/tests/general-helper-functions-test.metta``` you'll see that it passes all the test cases

Closes #204

### DIFF
--- a/pyvenv.cfg
+++ b/pyvenv.cfg
@@ -1,3 +1,0 @@
-home = /usr/bin
-include-system-site-packages = false
-version = 3.10.12
--- a/pyvenv.cfg
+++ b/pyvenv.cfg
@@ -1,0 +1,3 @@
+home = /usr/bin
+include-system-site-packages = false
+version = 3.10.12
--- a/utilities/general-helpers.metta
+++ b/utilities/general-helpers.metta
@@ -9,7 +9,8 @@
 
 ;Function to cocatinate two tuples (A B) (C D) ==> (A B C D)
 ;; (: concatTuple (-> Expression Expression Expression))
- (= (concatTuple $x $y) (collapse (union (superpose $x) (superpose $y)))) ;; FIX: No longer working on version 0.2.3
+;  (= (concatTuple $x $y) (collapse (union (superpose $x) (superpose $y)))) ;; FIX: No longer working on version 0.2.3
+(= (concatTuple $x $y) (union-atom $x $y)) ;; This should work when run after activating the virtual env
 (: ++ (-> Expression Expression Expression))
 (= (++ $x $y)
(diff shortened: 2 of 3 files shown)


utilities/general-helpers.metta
```diff
@@ -9,7 +9,8 @@
 
 ;Function to cocatinate two tuples (A B) (C D) ==> (A B C D)
 ;; (: concatTuple (-> Expression Expression Expression))
- (= (concatTuple $x $y) (collapse (union (superpose $x) (superpose $y)))) ;; FIX: No longer working on version 0.2.3
+;  (= (concatTuple $x $y) (collapse (union (superpose $x) (superpose $y)))) ;; FIX: No longer working on version 0.2.3
```
Yagth, line 12: Remove this line. It's no longer needed.

pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = /usr/bin
```
Yagth, line 1: Is this file necessary? If not could you remove it?

utilities/general-helpers.metta
```diff
@@ -9,7 +9,8 @@
...
+(= (concatTuple $x $y) (union-atom $x $y)) ;; This should work when run after activating the virtual env
```
Yagth, line 13: I don't think `union-atom` will be a permanent fix. Can you use a recursive definition for it instead?

utilities/general-helpers.metta
blackhammer116: Done!

Yagth (approved):

PR #219, opened 2025-04-07
Fix mentioned issues

### LINKED ISSUE #205: `hasOneAtom` helper function stopped working
**Description**

The function `hasOneAtom` found in the `general-helpers.metta` file doesn't work on the new version MeTTa on certain cases. The function returns `False` on the input `(hasOneAtom (foo))` while it should've returned `True` instead.

**Your Environment**

* Version used:
* MeTTa version (e.g. 0.1.11):0.2.3

_After completing the issue, make sure to run the commented out test case found in the `utilities/tests/general-helper-functions-tests.metta` file corresponding to the function._

### LINKED ISSUE #206: zip function implementation
**Detailed Description**

The zip function currently doesn't work since the `hasOneAtom` function doesn't behave as expected. However, the function could be reimplemented using the `zipWith` function. This reduce code redundancy and fix the issue at the same time. After re implementing it, the commented out test cases of the `zip` function found in `utilities/tests/general-helper-functions-tests.metta` file.

_After completing the issue, make sure to run the commented out test case found in the `utilities/tests/general-helper-functions-tests.metta` file corresponding to the function._

### LINKED ISSUE #207: Append atom doesn't work
**Description**

The `appendAtom` function totally stopped working on the new MeTTa version. It always returns a bad type error.
E.g: `(appendAtom 1 (1 2))`

**Expected Behavior**

The above code's expected result was: `(1 2 1)`

**Actual Behavior**

However the code is returning: `(Error 1 BadType)`.

* Version used:
* MeTTa version: 0.2.3

_After completing the issue, make sure to run the commented out test case found in the `utilities/tests/general-helper-functions-tests.metta` file corresponding to the function._

### LINKED ISSUE #208: Len doesn't work on a symbol.
**Detailed Description**

The `len`function should return more useful error when called on a symbol. This would during debugging and avoids confusing the `len` function's error with another error. 

_After completing the issue, make sure to run the commented out test case found in the `utilities/tests/general-helper-functions-tests.metta` file corresponding to the function._

yotors:
## Description
This PR addresses several issues and improves the reliability and type safety of general helper functions within Metta 0.2.3. Several failing test cases have been resolved, and the implementation of several functions has been refined.

## Motivation and Context

**Key Changes:**

* **Improved Handling of Non-Determinism:**
    * ND tests now use equality (`===`) to accommodate non-deterministic results.
    * `setDifference` implements a non-deterministic approach using subtraction. Tests have been updated to account for this. (Order is assumed to be irrelevant; `subtraction-atom` can be considered if order matters.)
    * `findAndReplace` test cases now handle non-deterministic outcomes.

* **Bug Fixes:**
    * `hasOneAtom` corrected to check both atom type and length, resolving previous issues. Closes #205 
    * `until` function type notation corrected.
    * `filter` test cases fixed by addressing related issues in `hasOneAtom` and type notation. (This fixes compatibility issues with stricter type handling in newer Metta versions.)
    * `update-atom` The implementation only deletes the first occurrence of a given atom, but the test assumes the deletion of all occurrences. So, I fixed the test. (We may use ND to delete all of its occurrences if needed) 
    * `gen` function issue is resolved by assigning the reduction of `(car-atom)` to a variable before conditional checks.

* **Implementation Enhancements:**
    * `append` correct type notation to use different variable for different types. Closes #207 
    * `zip` base case improved to handle empty lists, preventing errors. I tried to use zipWith using cons-atom as a function, but cons-atom needs Expression as a second parameter. Closes #206 
    * `len` now returns an error for non-expression atoms. Closes #208 

## How Has This Been Tested?
* **Test Coverage:** All test cases for general helper functions are now passing in Metta 0.2.3. This addresses several previously reported issues.

## Types of changes

## Checklist:

### DIFF
--- a/utilities/general-helpers.metta
+++ b/utilities/general-helpers.metta
@@ -41,20 +41,20 @@
 ; The function uses isMember from general helpers functions.
 
 (: setDifference (-> Expression Expression Expression))
-(= (setDifference $set1 $set2) (collapse (subtraction (superpose $set1) (superpose $set2))))
-    ; (if (== $set1 ()) () ; Base case: return empty set when $set1 is empty
-    ;     (let* (
-    ;         ($head (car-atom $set1))  ; Extract the head of set1
-    ;         ($tail (cdr-atom $set1))  ; Extract the tail of set1
-    ;         ($newtail (setDifference $tail $set2))  ; Recursively process the tail
-    ;     )
-    ;         ; Check if the $head is in set2, if so, skip it
+(= (setDifference $set1 $set2) 
+    (if (== $set1 ()) () ; Base case: return empty set when $set1 is empty
+        (let* (
+            ($head (car-atom $set1))  ; Extract the head of set1
+            ($tail (cdr-atom $set1))  ; Extract the tail of set1
+            ($newtail (setDifference $tail $set2))  ; Recursively process the tail
+        )
+            ; Check if the $head is in set2, if so, skip it
 
-    ;         (if (isMember $head $set2)
-    ;             $newtail 
-    ;             ; Otherwise, add the head to the new tail, avoiding unnecessary parentheses
-    ;             (if (== (get-metatype $newtail) Expression) (cons-atom $head $newtail) $head)  ; Avoid wrapping the head if it's already properly structured
-    ;         ))))
+            (if (isMember $head $set2)
+                $newtail 
+                ; Otherwise, add the head to the new tail, avoiding unnecessary parentheses
+                (if (== (get-metatype $newtail) Expression) (cons-atom $head $newtail) $head)  ; Avoid wrapping the head if it's already properly structured
+            ))))
 
 ;; A helper function for the findAndReplace function.
 ;; It handles the replacement of a single atom.
@@ -231,13 +231,8 @@
          (cons-atom $head $tail))))
 
 ;; Finds length of a tuple 
-(= (len $expr)
-    (if (not (isExpression $expr)) (Error $expr "is not Expression")
-    (if (== $expr ())
-        0
-        (let $tail (cdr-atom $expr)
-            (+ 1 (len $tail))
-        ))))
+(= (len $expr) (size-atom $expr))
+    
 
 ;; A function to replace an atom at a specific index position with $val -- 0 indexed counting is implied
 (: replaceByIndex (-> Expression Number $t Expression))
--- a/utilities/general-helpers.metta
+++ b/utilities/general-helpers.metta
@@ -79,7 +79,7 @@
 
 ;; A function that behaves like a do while loop.
 ;; It executes the functions in the given list and returns the last result.
-(: until (-> (-> $a Bool) (-> $b $b) $b $b))
+(: until (-> (-> $a Bool) (-> $a $a) $a $a))
 (= (until $predicate $function $x) (if ($predicate $x) $x (let $x' ($function $x) (until $predicate $function $x'))))
 
 ;; A foldleft function like the implementation in Haskell.
@@ -175,7 +175,7 @@
 
 ;; A function to check if a given atom is an expression 
 ;;   with has one atom in it.
-(: hasOneAtom (-> Atom Bool))
+(: hasOneAtom (-> $a Bool))
 (= (hasOneAtom $expr) (if (== (get-metatype $expr) Expression) (== (size-atom $expr) 1) False))
 
 ;; Function to replace a given atom from a space 
--- a/utilities/general-helpers.metta
+++ b/utilities/general-helpers.metta
@@ -41,20 +41,20 @@
 ; The function uses isMember from general helpers functions.
 
 (: setDifference (-> Expression Expression Expression))
-(= (setDifference $set1 $set2) 
-    (if (== $set1 ()) () ; Base case: return empty set when $set1 is empty
-        (let* (
-            ($head (car-atom $set1))  ; Extract the head of set1
-            ($tail (cdr-atom $set1))  ; Extract the tail of set1
-            ($newtail (setDifference $tail $set2))  ; Recursively process the tail
-        )
-            ; Check if the $head is in set2, if so, skip it
+(= (setDifference $set1 $set2) (collapse (subtraction (superpose $set1) (superpose $set2))))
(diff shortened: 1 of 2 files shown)


utilities/general-helpers.metta
```diff
@@ -41,20 +41,20 @@
...
-            ($newtail (setDifference $tail $set2))  ; Recursively process the tail
-        )
-            ; Check if the $head is in set2, if so, skip it
-
-            (if (isMember $head $set2)
-                $newtail 
-                ; Otherwise, add the head to the new tail, avoiding unnecessary parentheses
-                (if (== (get-metatype $newtail) Expression) (cons-atom $head $newtail) $head)  ; Avoid wrapping the head if it's already properly structured
-            ))))
+(= (setDifference $set1 $set2) (collapse (subtraction (superpose $set1) (superpose $set2))))
```
Yagth, line 44: I understand the change you've made will improve the code, but we need the `setDifference` function to work in an ordered manner and we wanted to define our own function just incase the built in versions fail to preserve order from update to update (We've seen this happen). There are functions which depend on it so it's best to keep it as it is for the time being.

utilities/general-helpers.metta
```diff
@@ -79,7 +79,7 @@
 
 ;; A function that behaves like a do while loop.
 ;; It executes the functions in the given list and returns the last result.
-(: until (-> (-> $a Bool) (-> $a $a) $a $a))
+(: until (-> (-> $a Bool) (-> $b $b) $b $b))
```
Yagth, line 82: This isn't a correct fix. The `$a` type variable is expected to be the same as the predicate's `$a` type variable.

utilities/general-helpers.metta
```diff
@@ -229,13 +230,14 @@
               ($tail (zipWith $f $xs-tl $ys-tl)))
          (cons-atom $head $tail))))
 
-;; Finds length of a tuple ;; FIX: No longer working on version 0.2.3
+;; Finds length of a tuple 
 (= (len $expr)
-    (if (== $expr ()) ;; FIX: return Error if symbol
+    (if (not (isExpression $expr)) (Error $expr "is not Expression")
+    (if (== $expr ())
```
Yagth, line 236: Why not use `size-atom` here too?

utilities/tests/general-helper-functions-test.metta
```diff
@@ -174,59 +174,60 @@
...
-;; ;; FIX: Failing Test case
-;; !(assertEqual (len (A B C D)) 4)
-;; !(assertEqual (len ()) 0)
-;; !(assertEqual (len (A)) 1)
-;; !(assertEqual (len A) 4) ;; Should return Error. !(assertEqual (replaceByIndex (A B C) 0 D) (D B C))
+;; ;; FIXED: Failing Test case
+!(assertEqual (len (A B C D)) 4)
+!(assertEqual (len ()) 0)
+!(assertEqual (len (A)) 1)
+!(assertEqual (len A) (Error A "is not Expression")) ;; Should return Error.
```
Yagth, line 228: Remove the trailing comment.

Yagth (changes requested):

utilities/general-helpers.metta
```diff
@@ -124,8 +124,8 @@
 (:wrapper (-> (-> $a Bool) $aa $aa))
 (= (wrapper $predicate $x) (if (let $bool ($predicate $x) $bool) $x (empty)))
 
-;; Filter function given a predicate ;; FIX: No longer working on version 0.2.3
-(: filter (-> (-> $a Bool) $a $a))
+;; Filter function given a predicate 
+(: filter (-> (-> $a Bool) $b $b))
```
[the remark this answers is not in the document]
yotors: But we are passing the superpose of an expression (which may contain non expression) to a predicate which may make $a to bind with non-expression which causes an error when we pass expression as the second parameter.

utilities/general-helpers.metta
yotors: Actually, we can. I thought you want to have ur own.

utilities/general-helpers.metta
yotors: Yeah ur correct. I thought we used superpose here. but 

` !(assertEqual (until isUnit cdr-atom (foo bar bae)) ())`
`!(assertEqual (until hasOneAtom cdr-atom (foo bar bae)) (bae)) ` 
was commented out and flagged as needing a fix. but the testes are already passing

yotors: The functions we used in the tests, like isUnit, which expects Atom as the first argument and cdr-atom, which expects Expression, cause errors. I first thought Atom is a super type for Expression, so that shouldn't be an error. But when I tested it using other simpler functions, it happened that the metta type treated them differently.

```
(: test (-> Atom Atom))
(= (test $x) 3)

(: test2 (-> Number Number))
(= (test2 $x) 4)

(: main (-> (-> $a $a) (-> $a $a) $a))
(= (main $f1 $f2) 5)

!(main test test2)
```
this return (Error test2 BadType)

utilities/tests/general-helper-functions-test.metta
yotors: k

utilities/general-helpers.metta
yotors: Tests for size-atom failed due to assertEqual evaluating the returned Error (Error (size-atom A) Atom is not an ExpressionAtom), causing an incorrect argument count error. Neither direct comparison nor assertEqualToResult (I don't know why) worked.  and also size-atom returns different error for symbol and Number. I just check if its return is equal to the return of size-atom.

utilities/general-helpers.metta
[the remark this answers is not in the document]
Yagth: Your're right. Okay lets keep this change. It makes sense to keep it like that.

utilities/general-helpers.metta
Yagth: That's weird, Error shouldn't have returned a entire sentence outside of a tuple. Anyways it's fine.

utilities/general-helpers.metta
Yagth: Don't forget to return the until function's type back to how it was :)

yotors: I forgot neber 😄

PR #222, opened 2025-04-08
MultiMap implementation

yotors:
## Description

This change introduces a `MultiMap` data structure that supports nested `MultiMap` values. The following methods have been implemented, all prefixed with `MultiMap.`:

* `MultiMap.insert`: Inserts a value associated with a key in order. The key can be a duplicate`.
* `MultiMap.findOne(key)`: Returns the first value (if any) associated with the key`.
* `MultiMap.findAll`: Returns an expression that contains all values associated with the key in a Map. Returns a unit atom if none exists. 
* `MultiMap.removeOne`: Removes the first value associated with the key and return the MultiMap.
* `MultiMap.removeAll`: Removes all values associated with the key and return the MultiMap.
* `MultiMap.contains`: Returns `true` if the key exists in the `MultiMap`, `false` otherwise.
* `MultiMap.checkValue`: Returns `true` if the specified value exists in the MultiMap, `false` otherwise.
* `MultiMap.length`: Returns the total number of key-value pairs in the `MultiMap`.

## Motivation and Context

This implementation addresses the need for a more versatile data structure than a standard map or multimap.

## How Has This Been Tested?

The `MultiMap` implementation has been tested using testscases

## Types of changes

## Checklist:

### DIFF
--- a/utilities/ordered-multimap.metta
+++ b/utilities/ordered-multimap.metta
@@ -1,57 +1,57 @@
 ;; Ordered multimap Data Structure
-(: OMM (-> ($k $v) Type))
-(: NilOMM (OMM ($k $v)))
-(: ConsOMM (-> ($k $v) (OMM ($k $v)) (OMM ($k $v))))
+(: MultiMap (-> ($k $v) Type))
+(: NilMMap (MultiMap ($k $v)))
+(: ConsMMap (-> ($k $v) (MultiMap ($k $v)) (MultiMap ($k $v))))
 
 ;; insert an element to the Ordered multimap
-(: OMM.insert (-> ($k $v) (OMM ($k $v)) (OMM ($k $v))))
-(= (OMM.insert ($key $value) NilOMM) (ConsOMM ($key $value) NilOMM))
-(= (OMM.insert ($key $value) (ConsOMM ($curKey $curVal) $tail))
+(: MultiMap.insert (-> ($k $v) (MultiMap ($k $v)) (MultiMap ($k $v))))
+(= (MultiMap.insert ($key $value) NilMMap) (ConsMMap ($key $value) NilMMap))
+(= (MultiMap.insert ($key $value) (ConsMMap ($curKey $curVal) $tail))
         (if (<= $key $curKey)
-            (ConsOMM ($key $value) (ConsOMM ($curKey $curVal) $tail))
-            (ConsOMM ($curKey $curVal) (OMM.insert ($key $value) $tail))))
+            (ConsMMap ($key $value) (ConsMMap ($curKey $curVal) $tail))
+            (ConsMMap ($curKey $curVal) (MultiMap.insert ($key $value) $tail))))
 
 ;; Get the first found value with the given key from the Ordered multimap using key
-(: OMM.findOne (-> $k (OMM ($k $v)) $v))
-(= (OMM.findOne $key NilOMM) (Error $key "not found"))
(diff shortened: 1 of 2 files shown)


utilities/ordered-multimap.metta
```diff
@@ -0,0 +1,57 @@
+;; Ordered multimap Data Structure
+(: MultiMap (-> ($k $v) Type))
+(: NilMMap (MultiMap ($k $v)))
+(: ConsMMap (-> ($k $v) (MultiMap ($k $v)) (MultiMap ($k $v))))
+
+;; insert an element to the Ordered multimap
+(: MultiMap.insert (-> ($k $v) (MultiMap ($k $v)) (MultiMap ($k $v))))
+(= (MultiMap.insert ($key $value) NilMMap) (ConsMMap ($key $value) NilMMap))
+(= (MultiMap.insert ($key $value) (ConsMMap ($curKey $curVal) $tail))
+        (if (<= $key $curKey)
```
Yagth, line 10: Can you overload the insert function so that it can accept a custom comparator function too instead of the hard coded `<=` only? This way if the user calls the insert function with their own `lte` function then that will be used but if they chose not to pass a compare function then the default `<=` will be used.

utilities/ordered-multimap.metta
```diff
@@ -0,0 +1,57 @@
...
+            (ConsMMap ($curKey $curVal) (MultiMap.insert ($key $value) $tail))))
+
+;; Get the first found value with the given key from the Ordered multimap using key
+(: MultiMap.findOne (-> $k (MultiMap ($k $v)) $v))
+(= (MultiMap.findOne $key NilMMap) (Error $key "not found"))
+(= (MultiMap.findOne $key (ConsMMap ($curKey $curVal) $tail))
+    (if (== $key $curKey) $curVal (MultiMap.findOne $key $tail) ))
+
+;; Get all found values with the given key from the Ordered multimap using key
+(: MultiMap.findAll (-> $k (MultiMap ($k $v)) Expression))
```
Yagth, line 21: We want to return a list instead of `expression`.

utilities/ordered-multimap.metta
yotors: Sure.

utilities/ordered-multimap.metta
yotors: Done.

utilities/ordered-multimap.metta
```diff
@@ -0,0 +1,63 @@
+;; Ordered multimap Data Structure
+(: MultiMap (-> ($k $v) Type))
+(: NilMMap (MultiMap ($k $v)))
+(: ConsMMap (-> ($k $v) (MultiMap ($k $v)) (MultiMap ($k $v))))
+
+;; default insert function that insert an element to the Ordered MultiMap using <=.
+(: MultiMap.insert (-> ($k $v) (MultiMap ($k $v)) (MultiMap ($k $v))))
+(= (MultiMap.insert ($key $value) NilMMap) (ConsMMap ($key $value) NilMMap))
```
Yagth, line 8: I think the above logic can get simplified if you use a single overload like so.
This will handle the default case, we don't need to define the other patterns :)

```suggestion
(= (MultiMap.insert $newValue $mMap) (MultiMap.insert $newValue $mMap <=))
```

yotors: Yeah 👍

Yagth (approved):

PR #224, opened 2025-04-12
update reduct - python implementation to handle metta expressions 

eyuuab:
## Types of changes

## Checklist:


Yagth:
@eyuuab did you test the import in a different folder structure like out of the reduct folder?

eyuuab:
yeah, it works like importing any other metta file

Yagth:
Awesome.  I will take a look on Monday.

eyuuab:
[Python Reduction Integration](https://trello.com/c/ZhpEkTyC/159-python-reduction-integration)

reduct/enf/DataStructures/Graphs.py
```diff
@@ -0,0 +1,18 @@
+from enum import Enum
```
Yagth, line 1: I don't think we need this file. You can remove it.

reduct/enf/Tests/EnfRuleCheckers.py
```diff
@@ -0,0 +1,316 @@
+from functools import reduce
```
Yagth, line 1: I don't think we need this file too.

reduct/enf/Utilities/TraverseGraph.py
```diff
@@ -0,0 +1,39 @@
+from ..DataStructures.Graphs import ConstraintGraphNode,GraphNodeType,NodeType
```
Yagth, line 1: We don't need this file.

reduct/enf/test.py
```diff
@@ -0,0 +1,29 @@
+from operator import itemgetter
```
Yagth, line 1: We don't need this file.

reduct/enf/Utilities/parser.py
```diff
@@ -0,0 +1,78 @@
...
+            return to_binary(args)
+
+        else:
+            raise ValueError(f"Unknown operator: {op}")
+
+    tokens = re.findall(r'\(|\)|[A-Za-z]+', expr)
+    return parse_tokens(tokens)
+
+## accepts constraint tree and returns metta boolean expression
+def parse_constraint_tree(text):
```
Yagth, line 42: This part could also be done in a simpler manner if only it accepts the tree node. In that case, it would have only been a matter of traversing the tree instead of parsing the complicated string representations of the constraint tree. But no need to refactor it now, just mentioned it FYI.

reduct/enf/Utilities/parser.py
```diff
@@ -0,0 +1,78 @@
...
+            
+            return to_binary(args)
+
+        else:
+            raise ValueError(f"Unknown operator: {op}")
+
+    tokens = re.findall(r'\(|\)|[A-Za-z]+', expr)
+    return parse_tokens(tokens)
+
+## accepts constraint tree and returns metta boolean expression
```
Yagth, line 41: Making it clear in the comment that the constraint tree mentioned here is a constraint tree's string representation would be a lot helpful.

reduct/enf/Utilities/parser.py
Yagth, line 42: FYI, this function would be a lot simpler if it accepted the constraint tree itself instead of it's string representation. That way, it would have only been a matter of traversing the tree.

reduct/enf/.gitignore
```diff
@@ -0,0 +1,2 @@
+__pycache__
```
Yagth, line 1: This `.gitignore` file is redundent, remove it from here.

Yagth (changes requested):
Everything else is looking good. Just remove the files mentioned as unnecessary and address few comments I mentioned. Apart from that, we also need to remove the "Tests" and the "alternative_implementation" folders. They're not necessary.

Yagth (approved):

PR #228, opened 2025-04-17
Fixed the bug on delete-inconsistent-handle file

blackhammer116:
## Description
The non-deterministic result did not keep the order of the elements in the tuple as it should, since it worked with older functions, thus I've built a recursive function ```reverseExp``` to reverse a single layer of elements inside the tuple. Also, the ```isConsistentExp```  function was failing when given a Symbol, which is now handled in a ```if```` clause.

## How Has This Been Tested?
Run the test cases inside the ```delete-inconsistent-tests``` file.


reduct/boolean-reduct/rte-helpers.metta
```diff
@@ -14,17 +14,21 @@
...
+      ()
+      (if (== (car-atom $exp) OR)
+          ()
+          (getLiterals $exp)))))
 
 ;; A function similar to the getGuardSet function but returns a
 ;;    non deterministic list instead of a collapsed atom.
 ;; Example: (AND (NOT A) (NOT B) A) -> [(NOT A), (NOT B), A]
 (= (getGuardSetND $exp)
+  (if (isSymbol $exp) (filterLiterals $exp)
```
Yagth, line 28: Same comment here. If the `$exp` is symbol there's no need to give it to the `filterLiterals` function.

reduct/boolean-reduct/delete-inconsistent-handle.metta
```diff
@@ -29,6 +29,8 @@
 ;; Example: input: (A B C), output: True
 ;;          input: (A B C (NOT A)), output: True
 (= (isConsistentExp $handleSet)
+  (if (isSymbol $handleSet) ;; added a symbol check
```
Yagth, line 32: Is there really a case where the `$handleSet` contains symobls?

utilities/general-helpers.metta
```diff
@@ -105,8 +105,8 @@
 (= (foldr $f $i $xs)
    (if (== $xs ())
        $i
-       (let* (($h (car-atom $xs))
-              ($t (cdr-atom $xs))
+       (let* (($h  (if (isSymbol $xs) $xs (car-atom $xs))) ; added isSymbol checks not to encounter any errors
```
Yagth, line 108: I think you can replace this logic using `decons-atom` and then we would only have one `isSymbol` check (If the check is really necessary).

Yagth: Did you encounter a test case or usecases where there is a `Symbol` being passed to it?

utilities/general-helpers.metta
```diff
@@ -498,3 +498,23 @@
...
+;;                              Eg: given ((AND E C) B A OR) => (OR A B (AND E C))
+;;                              NOTE: Must use collapse at the begining of your function call
+;; Usage: (collapse (reversExp ((AND E C) B A OR)))
+(= (reverseExp $exp $i)
+  (if (== $exp ())
+    $i
+    (let*
+      (
+        (($h $t) (decons-atom $exp))
+        ($s (if (isSymbol $h) (cons-atom $h $i) (++ ($h) $i)))
```
Yagth, line 515: This logic is redundent. Why don't you just use the `cons-atom` with the `$h` and the `$i`? Why is the `if` check necessary?

reduct/boolean-reduct/delete-inconsistent-handle.metta
blackhammer116: Yes, some of the tests will fail if there is no check

utilities/general-helpers.metta
blackhammer116: Yes, since car-atom can't process Symbols it will fail.

utilities/general-helpers.metta
blackhammer116: if an expression is passed, the expression is not kept; rather, it deconstructs the expression into a single tuple for example given ```++ (OR A B) (AND E F)``` this will result in ```(OR A B AND E F)``` rather than ```(OR A B (AND E F))``` that's the logic behind the if condition.

utilities/general-helpers.metta
Yagth: Yeah I understand that's the case but we we don't want to give the `foldr` function a symbol in the first place so it shouldn't handle it

utilities/general-helpers.metta
Yagth: Is there a test case that is specific to this kind of input? I tried using the example above but even with the inclusion of this logic it still returns `(OR A B AND E F)`

blackhammer116: yes this was the test case that was failing ```! (assertEqualToResult (deleteInconsistentHandle (OR A B (AND E A) (AND (NOT A) (NOT B) C)) (AND (NOT A) (NOT B) C) (F B G)) (((OR A B (AND E A)) () True)))
```
- The first logic was not to reduce the second layer of the tuple in this case ```(AND E A)``` when concatenating them.
- To solve that conundrum the way would be to and an extra ```()``` on the outside after checking its an expression.
- Thirdly since ```cons-atom``` doesn't expect an expression as the first argument it will fail and render a ```BadType``` Error
- Thus using ```++``` while adding the extra brackets solves this problem.

File: utilities/general-helpers.metta (hunk 3 above)
Comment (blackhammer116) [reply utilities/general-helpers.metta:108]: You're right, but when processing the expression, the tail started to become a Symbol rather than an expression so when it tries to select the head it renders an error indicating a Symbol was passed to ```car-atom``` rather than an expression. This is specific only when being implemented in the reverse function, under normal circumstances you're right ```foldr``` isn't supposed to process a Symbol

File: reduct/boolean-reduct/delete-inconsistent-handle.metta (hunk 2 above)
Comment (Yagth) [reply reduct/boolean-reduct/delete-inconsistent-handle.metta:32]: That's weird, because I tried it out commenting the `if` check for symbol and it all test cases pass.

File: utilities/general-helpers.metta (hunk 4 above)
Comment (Yagth) [reply utilities/general-helpers.metta:515]: I just ran all the `deleteInconsistentHandle` function by removing the `if` check and using `cons-atom` only instead of the `++` and all of the tests pass. Can you try it again locally?

**Comment by Yagth** [comment]
Few more comments:

1. I just noticed that there were instances in the `helper-function-test.metta` where you only used the `===` without an `assertEqual` function. The result of the `===` function should be wrapped in an `assertEqual` and it's expected result should either be `True` or `False` depending on the usecase. 

For example instead of the following:
```scheme
!(=== (getGuardSet (AND (NOT A) (NOT B) A)) ((NOT A) (NOT B) A))
```

You should do:

```scheme
!(assertEqual (=== (getGuardSet (AND (NOT A) (NOT B) A)) ((NOT A) (NOT B) A)) True)
```
2. You should rename the `delete-inconsistent-handle-tests.metta` file back to `delete-inconsistent-handle-test.metta` other wise, it won't be picked up by the test script.

Yagth (approved):

PR #231, opened 2025-04-19
Clean tree implementation

yotors:
Clean combo tree function,  .values, .keys, and  .items 

## Description

Clean combo tree function that takes a tree, reduces it, and returns the reduced tree.

Besides, I added .values, .keys, and  .items functions to the map and multimap structures. 

## How Has This Been Tested?

All the changes have passed

## Types of changes

## Checklist:

### DIFF
--- a/utilities/map.metta
+++ b/utilities/map.metta
@@ -39,6 +39,24 @@
 (= (Map.remove $key (ConsMap ($curKey $curVal) $tail)) 
     (if (== $key $curKey) $tail (ConsMap ($curKey $curVal) (Map.remove $key $tail))))
 
+;; Get the all the values from the map
+(: Map.values (-> (Map ($v $v)) Expression))
+(= (Map.values NilMap) ())
+(= (Map.values (ConsMap ($key $value) $tail))
+    (let $rest (Map.values $tail) (cons-atom $value $rest)))
+
+;; Get the all the keys from the map
+(: Map.keys (-> (Map ($v $v)) Expression))
+(= (Map.keys NilMap) ())
+(= (Map.keys (ConsMap ($key $value) $tail))
+    (let $rest (Map.keys $tail) (cons-atom $key $rest)))
+
+;; Get the all the key value pairs from the map
+(: Map.items (-> (Map ($v $v)) Expression))
+(= (Map.items NilMap) ())
+(= (Map.items (ConsMap $pair $tail))
+    (let $rest (Map.items $tail) (cons-atom $pair $rest)))
+
 ;; Get the length of the map
 (: Map.length (-> (Map ($k $v)) Number))
 (= (Map.length NilMap) 0)
--- a/utilities/ordered-multimap.metta
+++ b/utilities/ordered-multimap.metta
@@ -53,7 +53,27 @@
 (= (MultiMap.removeAll $key NilMMap) NilMMap)
 (= (MultiMap.removeAll $key (ConsMMap ($curKey $curVal) $tail)) 
     (if (== $key $curKey) (MultiMap.removeAll $key $tail) (ConsMMap ($curKey $curVal) (MultiMap.removeAll $key $tail))))
-    
+
+
+;; Get the all the values from the map
+(: MultiMap.values (-> (MultiMap ($v $v)) Expression))
+(= (MultiMap.values NilMMap) ())
+(= (MultiMap.values (ConsMultiMap ($key $value) $tail))
+    (let $rest (MultiMap.values $tail) (cons-atom $value $rest)))
+
+;; Get the all the keys from the MultiMap
+(: MultiMap.keys (-> (MultiMap ($v $v)) Expression))
+(= (MultiMap.keys NilMMap) ())
+(= (MultiMap.keys (ConsMultiMap ($key $value) $tail))
+    (let $rest (MultiMap.keys $tail) (cons-atom $key $rest)))
+
+;; Get the all the key value pairs from the MultiMap
+(: MultiMap.items (-> (MultiMap ($v $v)) Expression))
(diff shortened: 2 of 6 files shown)


Yagth:
[Clean Combo Tree](https://trello.com/c/x3TZRJBd/172-clean-combo-tree)

utilities/map.metta
```diff
@@ -39,6 +39,24 @@
 (= (Map.remove $key (ConsMap ($curKey $curVal) $tail)) 
     (if (== $key $curKey) $tail (ConsMap ($curKey $curVal) (Map.remove $key $tail))))
 
+;; Get the all the values from the map
+(: Map.values (-> (Map ($v $v)) Expression))
```
Yagth, line 43: As a convention, we prefer returning `(List $v)` instead of `Expression`. And also the type of map should not be `(Map ($v $v))`. That will force the map to have the same type for keys and values which isn't always the case.

utilities/map.metta
```diff
@@ -39,6 +39,24 @@
...
+(= (Map.values NilMap) ())
+(= (Map.values (ConsMap ($key $value) $tail))
+    (let $rest (Map.values $tail) (cons-atom $value $rest)))
+
+;; Get the all the keys from the map
+(: Map.keys (-> (Map ($v $v)) Expression))
```
Yagth, line 49: Same, here, instead of returning `Expression` We need to return `(List $k)` where `$k` is the type of the keys. And also correct the type of map here as well.

utilities/map.metta
```diff
@@ -39,6 +39,24 @@
...
+(= (Map.keys NilMap) ())
+(= (Map.keys (ConsMap ($key $value) $tail))
+    (let $rest (Map.keys $tail) (cons-atom $key $rest)))
+
+;; Get the all the key value pairs from the map
+(: Map.items (-> (Map ($v $v)) Expression))
```
Yagth, line 55: Same comment as in https://github.com/iCog-Labs-Dev/metta-moses/pull/231#discussion_r2053498594

utilities/ordered-multimap.metta
```diff
@@ -53,7 +53,27 @@
 (= (MultiMap.removeAll $key NilMMap) NilMMap)
 (= (MultiMap.removeAll $key (ConsMMap ($curKey $curVal) $tail)) 
     (if (== $key $curKey) (MultiMap.removeAll $key $tail) (ConsMMap ($curKey $curVal) (MultiMap.removeAll $key $tail))))
-    
+
+
+;; Get the all the values from the map
+(: MultiMap.values (-> (MultiMap ($v $v)) Expression))
```
Yagth, line 59: You should follow the same convetion here as well like in the previous file, https://github.com/iCog-Labs-Dev/metta-moses/pull/231#discussion_r2053498594

utilities/tests/ordered-multimap-test.metta
```diff
@@ -40,6 +40,22 @@
 ! (assertEqual (MultiMap.checkValue d (ConsMMap (1 a) (ConsMMap (2 b) (ConsMMap (3 c) (ConsMMap (4 d) NilMMap))))) True)
 ! (assertEqual (MultiMap.checkValue f (ConsMMap (1 a) (ConsMMap (2 b) (ConsMMap (3 c) NilMMap)))) False)
 
+;; Testcase for Map.values
+! (assertEqual (MultiMap.values (ConsMultiMap (1 a) (ConsMultiMap (2 b) (ConsMultiMap (3 c) NilMMap)))) (a b c))
```
Yagth, line 44: When building maps, we use `ConsMMap` as a constructor instead of the `ConsMultiMap` so change it accordingly.

utilities/tests/tree-test.metta
```diff
@@ -5,7 +5,10 @@
 ! (import! &self metta-moses:utilities:tree)
 ! (import! &self metta-moses:utilities:nodeId)
 ! (import! &self metta-moses:utilities:general-helpers)
+! (import! &self metta-moses:reduct:enf)
 
+;; Helper function for clean tree
+(= (REDUCE $expr) (reduce $expr))
```
Yagth, line 11: Why do you need to define another helper function `REDUCE` when you can directly use the `reduce` itself ?

utilities/tree.metta
```diff
@@ -30,6 +30,15 @@
             (mkTree (mkNode $expr) Nil))
             )
 
+(: cleanTree (-> (Tree $a) (Tree $a)))
+(= (cleanTree (mkTree (mkNode $r) (Cons $x $xs)))
+    (let*
+        (
+          ($preordered (preOrder (mkTree (mkNode $r) (Cons $x $xs))))
+          ($reduced (REDUCE $preordered))
```
Yagth, line 38: You should use the `reduce` function directly instead of a `REDUCE` function which is defined in the test case file. You should either move the `REDUCE` function here if it's existence has some relevance or replace it's use by the `reduce` function.

utilities/tree.metta
```diff
@@ -30,6 +30,15 @@
             (mkTree (mkNode $expr) Nil))
             )
 
+(: cleanTree (-> (Tree $a) (Tree $a)))
+(= (cleanTree (mkTree (mkNode $r) (Cons $x $xs)))
```
Yagth, line 34: You don't need to deconstruct the tree here, you could just use a variable as `$tree` and use that in the preOrder function. That would reduce the complexity of the function.

Yagth (changes requested):
Awesome work. Just few correction to make and good to merge.

utilities/tree.metta
Yagth: I see, okay I get it know :)

Yagth (approved):

PR #232, opened 2025-04-22
Fixed one-constraint-subsumption file

blackhammer116:
## Description
- The file consisted of errors similar to those encountered with the delete-inconsistent-handle file, which is:
1. ```car-atom``` keeps rendering BadType Errors due to a) an empty expression being passed to it, and b) a Symbol being passed to it
-- So I fixed that with a simple if check for both
2. Since the ```removeElement``` function returns a non-deterministic order, it was causing a problem with the test cases, thus I called the ```reverseExp``` similarly to what was done in the delete-inconsistent-handle.


reduct/boolean-reduct/tests/n-ary-gather-junctors-tests.metta
```diff
@@ -2,17 +2,17 @@
 
 ! (import! &self metta-moses:utilities:general-helpers)
 ! (import! &self metta-moses:reduct:boolean-reduct:n-ary-gather-junctors)
-; ! (import! &self metta-moses:reduct:boolean-reduct:rte-helpers)
+! (import! &self metta-moses:reduct:boolean-reduct:rte-helpers)
 
 
 ;; Test 01
-!(=== (gatherJunctors (OR A B)) (OR A B))
+; !(=== (gatherJunctors (OR A B)) (OR A B))
```
Yagth, line 9: Why are these test cases of the gatherJunctors commented out?

reduct/boolean-reduct/tests/n-ary-gather-junctors-tests.metta
```diff
@@ -21,10 +21,12 @@
 ; )
 
 ;; Test 03
-!(assertEqual
-    (gatherJunctors (OR (OR (OR C D) B) A (AND F H (AND K L) V)))
-    (OR C D B A (AND F H K L V))
-)
+!(isSame (gatherJunctors (OR (OR (OR C D) B) A (AND F H (AND K L) V)))
```
Yagth, line 24: I think this should be wrapped in an `assertEqual`

reduct/boolean-reduct/tests/n-ary-gather-junctors-tests.metta
Yagth: And if you're done with the gatherJunctors file refactor already, I think you should change the test file into an `*-test.metta`

reduct/boolean-reduct/tests/n-ary-gather-junctors-tests.metta
blackhammer116: That file is not done but will do once it is

blackhammer116:
Nope, it has nothing to do with it, its a separate task

reduct/boolean-reduct/tests/n-ary-gather-junctors-tests.metta
Yagth: Best if you remove them. I can't see the changes made if the code changes but is incomplete :)

Yagth: Or just change them back to how they were before :)

Yagth: If you're done with the `gatherJunctors` file then why isn't the test case renamed to the `-test.metta` file 🤔 ?

blackhammer116: Done!

Yagth (approved):

PR #235, opened 2025-04-23
fix: minor change on lazy Random Selector

eyuuab:
## Types of changes

## Checklist:

### DIFF
--- a/utilities/lazy-random-selector.metta
+++ b/utilities/lazy-random-selector.metta
@@ -29,7 +29,7 @@
     (chain (pick $lower $upper $picked ) $rand 
         (if  (== $n 0) 
             $picked
-            (if  (== (+ (- $upper $lower) 1) (length $picked))
+            (if  (< (+ (- $upper $lower) 1) (+ (size-atom $picked) $n))
                 (Error break "number of elements to select is greater than list size")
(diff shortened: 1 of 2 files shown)


utilities/lazy-random-selector.metta
```diff
@@ -27,10 +27,12 @@
 (= (lazyRandomSelector $lower $upper $n) (lazyRandomSelector $lower $upper $n ()))
 (= (lazyRandomSelector $lower $upper $n $picked)
     (chain (pick $lower $upper $picked ) $rand 
-            (if (or (== $n 0) (== (- $upper $lower) (length $picked)))
-                $picked
+        (if  (== $n 0) 
+            $picked
+            (if  (== (+ (- $upper $lower) 1) (length $picked))
```
Yagth, line 32: Hey, this logic works fine but it could be simplified right? If we're computing the length of the expression anyway, why not reduce the running time of the function by making sure nothing gets executed in the first place if the `((length $picked) + $n)` is always less or equal to the `(($upper - $lower) + 1)`? When I ran the code on inputs that are have higher difference between `$upper` and `$lower` with even higher `$n` the code takes a longer than it should to return an error as if it was still trying to computing something else. 
```suggestion
            (if  (<= (+ (- $upper $lower) 1) (+ (length $picked) $n))
```

Yagth (changes requested):

PR #247, opened 2025-05-08
Sampling from neighborhood

yotors:
## Description

The function returns sample size samples that are differ by dist specifications from the central instance.

## Motivation and Context

It returns a list of instances. I couldn't figure out how to use knobmapper instead of the field set in the CPP code.

## How Has This Been Tested?

Yes, it has been tested. But there are some points I doubt to be correct. Besides, the superposed type of the mkDiscKnob was causing bad type error. I commented out one.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My code follows the code style of this project.
- [X] I have added tests to cover my changes.


yotors:
It is creating a supposition of type, which causes an error in my function's (List mkDiscknob). I don't know why. 
```
! (bind! CENTRAL (Cons (mkDiscKnob (mkKnob (mkTree (mkNode B) Nil) (mkNodeId (3))) (mkMultip 2) (mkDiscSpec 0) (mkDiscSpec 0) Nil)(Cons (mkDiscKnob (mkKnob (mkTree (mkNode B) Nil) (mkNodeId (4))) (mkMultip 2) (mkDiscSpec 0) (mkDiscSpec 0) Nil) (Cons (mkDiscKnob (mkKnob (mkTree (mkNode A) Nil) (mkNodeId (5))) (mkMultip 2) (mkDiscSpec 0) (mkDiscSpec 0) Nil) Nil))))
! (get-type CENTRAL)
```
returned
```
[(List (-> Knob Multiplicity DiscreteKnob)), (List DiscreteKnob)]
```

Yagth:
> It is creating a supposition of type, which causes an error in my function's (List mkDiscknob). I don't know why.
> 
> ```
> ! (bind! CENTRAL (Cons (mkDiscKnob (mkKnob (mkTree (mkNode B) Nil) (mkNodeId (3))) (mkMultip 2) (mkDiscSpec 0) (mkDiscSpec 0) Nil)(Cons (mkDiscKnob (mkKnob (mkTree (mkNode B) Nil) (mkNodeId (4))) (mkMultip 2) (mkDiscSpec 0) (mkDiscSpec 0) Nil) (Cons (mkDiscKnob (mkKnob (mkTree (mkNode A) Nil) (mkNodeId (5))) (mkMultip 2) (mkDiscSpec 0) (mkDiscSpec 0) Nil) Nil))))
> ! (get-type CENTRAL)
> ```
> 
> returned
> 
> ```
> [(List (-> Knob Multiplicity DiscreteKnob)), (List DiscreteKnob)]
> ```
> 
> Thus (List.length CENTRAL) returns BadType error

That a weird error, we have been testing it for other functions and it was working fine so far. I guess the interpreter changes behavior on nested types. Any who, why did you have to create discKnobs only? We don't want to manipulate diskKnobs but LSKs. And it works fine if you change the constructor to an LSK. Or is it because of the task?

yotors:
The CPP implementation is on discrete set generally, but I think LSK will be enough, since we are focused on booleans.  I will change it to that. But why are we bulding disknobs in first place if we are only using LSK? 🤔

Yagth:
> The CPP implementation is on a discrete set generally, but I think LSK will be enough, since we are focused on booleans. I will change it to that. But why are we building disknobs in the first place if we are only using LSK? 🤔

We use the discKnobs as a way to abstract a commonly shared attributes between different kinds of knobs one of which is the LSK. But we could have another knobs like action knobs and the like in the future which will all share similar attributes as the discKnobs. However, we won't be using the discKnob directly. In the C++ version, there's a built in polymorphism in which one can convert a given knob type to it's parent whenever we need to have a generic function that works for all child knobs of discKnobs.

Yagth:
> I don't know how I can use knobmapper here (the CPP uses a field set to do some memory-level stuff). I just named the first parameter, which I am using it to keep the changes on, knobmapper. It has nothing to do with the knobmapper.

The C++ version takes the field set to locate the index of certain types of knobs and their corresponding specification from the instance. For example, it treats continues knobs, and bit knobs differently than it treats the discKnobs. But we only have discknobs so we don't need that logic.

The C++ version also have no way of locating where a certain knob is and how that relates to the instance without the use of fieldset because everything will be stored in the fieldset as a set of bits for memory efficiency. However, we can do the same thing using the knob mapper. We can find the knobs from the knob mapper and use their index in the knob mapper to find their corresponding specification in the instance. 

The C++ version updates the instance as a way of generating a new instance. However, in our case, it may be easier to just generate new instances.

You can perhaps see the implementation of `generateAllInNeighborhood` in the branch `feat/varyNKnobs` to see how we implemented a function similar to yours.

yotors:
> > I don't know how I can use knobmapper here (the CPP uses a field set to do some memory-level stuff). I just named the first parameter, which I am using it to keep the changes on, knobmapper. It has nothing to do with the knobmapper.
> 
> The C++ version takes the field set to locate the index of certain types of knobs and their corresponding specification from the instance. For example, it treats continues knobs, and bit knobs differently than it treats the discKnobs. But we only have discknobs so we don't need that logic.
> 
> The C++ version also have no way of locating where a certain knob is and how that relates to the instance without the use of fieldset because everything will be stored in the fieldset as a set of bits for memory efficiency. However, we can do the same thing using the knob mapper. We can find the knobs from the knob mapper and use their index in the knob mapper to find their corresponding specification in the instance.
> 
> The C++ version updates the instance as a way of generating a new instance. However, in our case, it may be easier to just generate new instances.
> 
> You can perhaps see the implementation of `generateAllInNeighborhood` in the branch `feat/varyNKnobs` to see how we implemented a function similar to yours.

Nice of you, thanks. I will take a look

representation/tests/sampling-from-neighborhood-test.metta
```diff
@@ -0,0 +1,63 @@
...
+              (mkDiscSpec 1)
+              (mkDiscSpec 1)
+              Nil)
+            (mkTree (mkNode A) Nil)))
+
+! (bind! knobMapObj (mkKbMap
+                      (ConsMap ((mkNodeId (2 3)) 0) (ConsMap ((mkNodeId (3)) 1) (ConsMap ((mkNodeId (1)) 2) NilMap)))
+                      (ConsMMap ((mkDiscSpec 1) lsk1) (ConsMMap ((mkDiscSpec 0) lsk2) (ConsMMap ((mkDiscSpec 1) lsk3) NilMMap)))))
+
+! (assertEqual (let $returned (sfn 2 1  knobMapObj (mkInst (Cons 1 (Cons 2 (Cons 3 Nil))))) (List.length $returned)) 2)
```
Yagth, line 60: This test case could be made more meaningfull if you also check that all the new instances generated have a distance of 1 from the center instance. I think there's already a built function to compute the hamming distance if you want to use that :). (Function is found in the neighborhood-sampling.metta file)

representation/sampling-from-neighborhood.metta
```diff
@@ -0,0 +1,53 @@
...
+(= (updateInst $dist $len $disc $newinst)
+    ( if (== $dist 0) 
+        $newinst
+        (let* (
+            (($rand) (lazyRandomSelector 0 (- $len 1) 1)) 
+            (($decspec (mkLSK (mkDiscKnob $knob (mkMultip $multi) $spec $spec1 $speclist) $tree)) (MultiMap.getByIdx $rand $disc))
+            ($curspec (List.getByIdx $newinst $rand))
+            ($upper (- $multi 1))
+            ($randomspec (+ (random-int &rng 0 $upper) 1)) 
+            ($updatedinst  (if (== $randomspec $curSpec)
```
Yagth, line 21: There's a type here. `$curSpec` should have been `$curspec`

PR #250, opened 2025-05-12
createDeme and countNeighborhood implementation

Liya-F:
## Description
1. Implemented the createDeme method.
2. Implemented the count_neighborhood_size_from_index and count_neighborhood_size methods found in the old moses codebase. It is used to count instances found at a specific distance from a given instance. 

## How Has This Been Tested?
I have tested it in metta 0.2.3 environments and it works fine. The change doesn't affect other parts of the codebase. 

## Types of changes

## Checklist:


optimization/hillclimbing/test/hillclimbing-helpers-test.metta
```diff
@@ -46,3 +127,21 @@
...
+  12)
+
+;Testcase for countNeighborhood
+!(assertEqual 
+ (countNeighborhood
+  knobmap3 ;knobmap
+  (mkInst (Cons 1 (Cons 0 (Cons 2 Nil)))) ;instance
+  3 ;distance
+  20) ;maxCount
+  8)
```
Yagth, line 147: Shouldn't the neighborhood count increase when the distance is increased? It's weird that it's doing the opposite.

Yagth:
[Deme Creation](https://trello.com/c/mfWHK4lK/116-deme-creation)

optimization/hillclimbing/test/hillclimbing-helpers-test.metta
Liya-F: yeah, but for example, let's have this instance [1 0 2], and multiplicity is always 3. If the distance is 2, 2 fields will only have 2 other possibilities, and the 3rd field will have all 3 possibilities. 2x2x3 = 12.

But when the distance is 3, all the fields can have 2 possibilities each. 2x2x2 = 8

Yagth: Okay I think I get it know. But can you also add this test case? I found it on the C++ version. This is how they test that the functions are working right. You can maybe add two test cases with different distances.
```scheme
!(let*
    (
       ($totalSize (countNeighborhood knobmap3 (mkInst (Cons 0 (Cons 0 (Cons 0 Nil)))) 3 20))
       ($generatedSize (List.length (generateAllInNeighborhood (mkMultip 3) (mkInst (Cons 0 (Cons 0 (Cons 0 Nil)))) 3)))
    )
    (<= (* 3 $totalSize) (* 4 $generatedSize)))
```

deme/create-deme.metta
```diff
@@ -0,0 +1,20 @@
...
+;; Return: A new empty deme with the specified representation and ID
+(: createDeme (-> (Tree $a) Number Number Deme)) 
+(= (createDeme $exemplar $nExpansion $nDeme)
+   (let*
+   (
+    ($demeIds (createDemeIds $nExpansion $nDeme))
+    ($representation (createRepresentation $exemplar))
+    ($demeId (List.getByIdx $demeIds 0))
+   )
+   (mkDeme $representation (mkSInstSet 0 Nil) $demeId)))
```
Yagth, line 20: Although this will be changed later with a more concrete score. It makes sens to make the score `-inf` than just assigning 0. In our implementation 0 is the highest score not the lowest. You can use the following to assign it the worst score possible. 

```scheme
(py-atom "float('-inf')")
```

Liya-F: I tried -inf as the starting score, and it worked fine, but for some reason, it couldn't pass the test case even if the 'Expected' and 'Got' of the check are literally the same. So, I replaced it with a large negative number, and it's working right this time. can we keep it this way?

Yagth: Well let's keep it like that for now. We got no other choice.

Yagth (approved):

PR #257, opened 2025-05-16
Bug fixes and additional test cases for the sample from neighborhood function

Yagth:


Yagth:
[Sample from neighborhood](https://trello.com/c/6p9ro2cd/156-sample-from-neighborhood)

PR #258, opened 2025-05-16
feat: implemented insersionSort helper function

yotors:
…pression

## Description

A simple insertion sort to partially sort an expression
## Motivation and Context

Sometimes we just need to sort the first $n elements from the expression to be sorted
## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My code follows the code style of this project.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/utilities/tests/general-helper-functions-test.metta
+++ b/utilities/tests/general-helper-functions-test.metta
@@ -249,6 +249,12 @@
 ;; !(assertEqual (sort (2 3 1 4) 4) (1 2 3 4))
 ;; !(assertEqual (sort (2 3 1 4) 4 >=) (4 3 2 1))
 
+!(assertEqual (insertionSort (9 7 5 2 4 6 3) 3 >=) (5 7 9 2 4 6 3))
+!(assertEqual (insertionSort (3 1 4 2 4 6 3) 5 <=) (4 4 3 2 1 6 3))
+!(assertEqual (insertionSort (9 7 5 2 4 6 3) 4) (2 5 7 9 4 6 3))
+!(assertEqual (insertionSort (4 7 5 2 4 6 3) <) (7 6 5 4 4 3 2))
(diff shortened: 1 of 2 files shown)


utilities/tests/general-helper-functions-test.metta
```diff
@@ -249,6 +249,12 @@
 ;; !(assertEqual (sort (2 3 1 4) 4) (1 2 3 4))
 ;; !(assertEqual (sort (2 3 1 4) 4 >=) (4 3 2 1))
 
+!(assertEqual (insertionSort (9 7 5 2 4 6 3) 3 >=) (5 7 9 2 4 6 3))
+!(assertEqual (insertionSort (3 1 4 2 4 6 3) 5 <=) (4 4 3 2 1 6 3))
+!(assertEqual (insertionSort (9 7 5 2 4 6 3) 4) (2 5 7 9 4 6 3))
+!(assertEqual (insertionSort (4 7 5 2 4 6 3) <) (7 6 5 4 4 3 2))
+!(assertEqual (insertionSort (4 7 5 2 4 6 3)) (2 3 4 4 5 6 7))
```
Yagth, line 256: For this test case for example, the parital sort should return the top 3 smallest element sorted in their order while leaving the rest of the list untouched.

Therefore the expected outcome when sorting the `(5 7 9 2 4 6 3)` in acsending order only `3` times would be `(2 3 4 ...)`. 

Another example would be, the expected outcome when sorting the `(5 7 9 2 4 6 3)` in descending order only `3` times would be `(9 7 6 ...)`.

Yagth (changes requested):

utilities/tests/general-helper-functions-test.metta
yotors: I tought it to sort the first $n elements of the expr rather that the first $n small elements. I will corect it then

PR #260, opened 2025-05-17
Fix for nary-propagate-not

blackhammer116:
## Description
Fixed the bug temporarily by using a recursive implementation. The nary-gather-junctors file's recursive implementation is still pending, am trying to look for work around but even the useful functions in stdlib are not working in this version.


Yagth (approved):

Yagth:
Don't forget about the new functions in metta like `map-atom`,  `foldl-atom`  and the like. Maybe they could be useful as a workaround instead of superposition

PR #271, opened 2025-05-31
Added Map.insetCounter function to insert elements into a map, updati…

Biruk-gebru:
### Description

Added the `Map.insetCounter` function, which inserts elements into a map by updating their counts based on occurrence frequency. This enhances the map's ability to track how many times each key appears, facilitating efficient counting operations. Additionally, test cases have been included to verify the correctness of this new functionality.

---

### Motivation and Context

This change provides a convenient way to perform counting within maps, streamlining operations that require tracking the number of occurrences of elements. It is useful for scenarios such as frequency analysis, histogram creation, and multiset implementations.

---

### How Has This Been Tested?

Test cases have been added to cover various scenarios, including inserting new elements, updating existing counts, and ensuring the correct counts are maintained after multiple insertions. Tests were run in the development environment to verify that the function behaves as expected and does not introduce regressions.

---

### Types of changes

- [x]  New feature (adds functionality)
- [ ] Bug fix (non-breaking change)
- [ ]  Breaking change (fix or feature that causes existing functionality to change)

---

### Checklist:

- [x]  My code follows the project's coding style.
- [x] All relevant tests passed.

### DIFF
--- a/utilities/map.metta
+++ b/utilities/map.metta
@@ -14,14 +14,14 @@
 
 ;; An insertion function that also allows map to act as a counter
 
-(: Map.insertCounter (-> ($k $v) (Map ($k $v)) (Map ($k $v))))
-(= (Map.insertCounter ($key $value) NilMap) (ConsMap ($key $value) NilMap))
-(= (Map.insertCounter ($key $value) (ConsMap ($curKey $curVal) $tail))
+(: Map.insertCounter (-> $k (Map ($k $v)) (Map ($k $v))))
+(= (Map.insertCounter $key NilMap) (ConsMap ($key 1) NilMap))
+(= (Map.insertCounter $key (ConsMap ($curKey $curVal) $tail))
     (if (== $key $curKey)
-        (ConsMap ($curKey (+ $curVal $value)) $tail)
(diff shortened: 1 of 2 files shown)


utilities/map.metta
```diff
@@ -12,6 +12,17 @@
             (ConsMap ($key $value) (ConsMap ($curKey $curVal) $tail))
             (ConsMap ($curKey $curVal) (Map.insert ($key $value) $tail)))))
 
+;; An insertion function that also allows map to act as a counter
+
+(: Map.insertCounter (-> ($k $v) (Map ($k $v)) (Map ($k $v))))
```
Yagth, line 17: When using the counter, we want to give it only the key. If the key exists in the map, then we increment the value. If the key isn't found in the map we simply add it with a count of 1. Can change your code to reflect that change?

Yagth (changes requested):

utilities/map.metta
Biruk-gebru: I've applied the requested changes. But wouldn't having a function to increment the value by more than one in a single call be helpful?

Yagth (approved):

Issue #256, opened 2025-05-15
Integrating python tests into github workflow


Yagth:
**Detailed Description**

After a while we have started using certain python functions integrated into MeTTa. These functions are written for performance gains. These functions also have their own python test cases but aren't being picked up neither by our github workflow nor by our testing script. Even though they are tested indirectly by the MeTTa functions, we want to have a separate run for the python tests as well.

Therefore, the task would be to extract these test cases from the main logic of the python code (if not already done so) and enable the workflow to identify and pick up these python tests and run them in addition to the MeTTa tests.

**Context**

Making this enables us to identify the cause of an error. (Whether it is from the error in the python file or a weird bug in MeTTa).

PR #276, opened 2025-06-10
Done with python helper function tasks

blackhammer116:
## Description
- Wrote a function that converts list to expression as instructed inside the listMethods file ```utilities/python-helpers/listMethods```
- Implemented a pytest test cases for both ```listMethods``` and ```treeMethods``` python functions
- Also edited the workflow .yaml file to automatically test these functions using pytest

Closes: #256 issue


blackhammer116:
the __init__.py function was creating an issue with pytest. pytest renders an error if the __init__.py file is present

blackhammer116:
Done with the task

utilities/python-helpers/test_python_helpers.py
```diff
@@ -0,0 +1,25 @@
+import unittest
+from .listMethods import exprToList, listToExpr
+
+
+class TestListMethods(unittest.TestCase):
+    
+    def test_listToExpr(self):
+        expected_result_1 = "(A B C)"
+        received_result_1 = listToExpr("(Cons A (Cons B (Cons C Nil)))")
+        expected_result_2 = "((mkInst (0 1 2)) (B) (C) D E)"
```
Yagth, line 10: This shouldn't have been the expected result if there are other datastructures inside. The function should only work on the top level list without changing the structure of any thing else that is an element of the list.

utilities/python-helpers/listMethods.py
```diff
@@ -44,6 +46,69 @@ def tokenize(expr: str):
...
+
+def listToExpr(list_str: str) -> str:
+    """
+    Main function to convert a Lisp-style cons list into a
+    formatted single-line string.
+    """
+    if not list_str.strip():
+        return "()"
+        
+    tokens = re.findall(r'\(|\)|[^\s()]+', list_str)
```
Yagth, line 104: I think this regex will cause a problem if there are nested list datastructures with in the first list

Yagth (changes requested):

utilities/python-helpers/test_python_helpers.py
Yagth: Yeah, that's what I meant. It should only work top level and not change any other nested datastructure.

.github/workflows/MeTTa-CI.yml
```diff
@@ -35,4 +35,4 @@ jobs:
     
     - name: Run Python tests
       working-directory: ./
-      run: python -m unittest discover -s utilities/
```
Yagth: I thought the `unittest` could be run to discover every test cases on it's own? Why don't you use that?

run_python_tests.py
```diff
@@ -0,0 +1,28 @@
+# run_tests.py
```
Yagth, line 1: I think the `unittest` command can handle this on it's own. If that's the case, we won't need this file.

Yagth (changes requested):

.github/workflows/MeTTa-CI.yml
Yagth: Let's do that for now. I wonder if we can find a better way in the future.

Yagth (approved):

PR #277, opened 2025-06-11
Optimization of sampleFromNeighborhood

yotors:
## Description

I used map-atom for the function to run sampleSize time instead of recursion. And used foldl-atom instead of recursion in the updateInst helper function.
## Motivation and Context

The function was too slow. So, I wanted to optimize it.

## How Has This Been Tested?

I didn't change the test file, and all the existing tests are passing
## Types of changes

- [ X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ X] My code follows the code style of this project.
- [ X] All new and existing tests passed.


Yagth (commented):
There are changes in your commit that we have already updated. I think it's best if you update your fork before sending another PR. That's probably why there are test cases failing, due to bugs that we already have fixed.

PR #330, opened 2025-08-24
hillclimbing for feature selection

Liya-F:
## Description
Starting from the already existing hillclimbing algorithm (which is used for deme optimization), I have created another for the process of feature selection since we are going to use it on both. I have made changes related to scoring, neighborhood exploration, crossing over and instance representation.

Instances are represented as a set of 1's and 0's representing the existence or absence of input features with the order as they appear in our input table. I found this really handy for the algorithm to apply the steps. Key things:
1. I have used Mutual information to score the instances. This is using the `bitsToIndices` function to first convert the list of 1's and 0's into a tuple of the features that are indicated to be present. Then I used `zipColumn` to zip the columns of these features and score it. 
2. On the neighborhood sampling part ,I have changed the neighborhood estimation logic to work for the representation we have now. `informationTheoreticBitsFs` uses the number of features we have. Also the `countNeighborhoodFromIdxFs` uses the binomial coefficient to count the number of instances at a specific distance from a given instance. `sampleNewInstancesFs` uses a multiplicity of 2 and the input table that the knob mapper. 
3. The final `hillClimbingFs` uses the updated methods specific to feature selection.

## Motivation and Context
This will be one of our feature selection algorithms that we are going to use in the future. The c++ codebase also uses the hill climbing as both a way to optimize a deme and to select the optimal feature set.  

## How Has This Been Tested?
Been tested locally inside a mettalog environment. 

## Types of changes

## Checklist:

### DIFF
--- a/moses/neighborhood-sampling.metta
+++ b/moses/neighborhood-sampling.metta
@@ -148,19 +148,47 @@
                               (List.replaceAt $newInst $rand 0)
                               (List.replaceAt $newInst $rand $randomSpec))))
 
-;; This function is Pre and post-processor function for sampleFromNeighborhood` 
-;; to achive not-deterministic approach. It generates a list with the size of $sampleSize 
-;; and calls sampleFromNeighborhood` function for every number in the list. 
-;; This will make the sampleFromNeighborhood` to produce $sampleSize samples. 
-;; Then the produced expression will be made in to llist using exprToList
+;; Pre/post-processor for `sampleFromNeighborhood` to ensure non-deterministic sampling.
+;; It checks if $sampleSize neighbors can be generated; if not, returns an error.
+;; Otherwise, it calls `selectUnique` to produce $sampleSize unique instances at the given distance.
+;; Parameters:
+;;      $sampleSize: Number of instances to generate
+;;      $dist: Distance from the center instance
+;;      $kbMp: Knob map for variations
+;;      $inst: Center instance
+;; Returns: A list of $sampleSize instances or an error if generation is not possible.
+
 (: sampleFromNeighborhood (-> Number Number KnobMap Instance (List Instance)))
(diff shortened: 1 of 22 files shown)


moses/neighborhood-sampling.metta
```diff
@@ -119,6 +137,18 @@
...
+(: updatedInstFs (-> Number Instance Instance))
+(= (updatedInstFs $rand $newInst)
+        (let*
+          (
+            ($curSpec (List.getByIdx $newInst $rand))
+            ($randomSpec (+ (random-int &rng 0 1) 1))
+            ($updatedInst (if (== $randomSpec $curSpec)
+                              (List.replaceAt $newInst $rand 0)
+                              (List.replaceAt $newInst $rand $randomSpec))))
+          $updatedInst))
```
K1rubel, line 149: I think you mean to name this `updateInstFs`. And would be better if you were to return the `if `block as a whole instead of `$updatedInst`.

optimization/hillclimbing/hill-climbing-helpers.metta
```diff
@@ -8,6 +8,12 @@
         (chain (List.foldr add (mkMultip 0) $multips) $res
           (let (mkMultip $val) $res $val)))))
 
+;; feature selection version of the informationTheoreticBits function
+;; In the case of feature selection, the information theoretic bits is simply the number of features we have
+(: informationTheoreticBitsFs (-> ITable Number))
+(= (informationTheoreticBitsFs (mkITable (Cons $row $rows) $labels))
```
K1rubel, line 14: Can it be sth like this? `(= (informationTheoreticBitsFs $labels) ..` since we are not making use of the rows part of the table. Obviously, the type definition also changes accordingly.

Yagth:
@Liya-F can you resolve the merge conflict? I was about to merge this PR

Yagth (approved):

PR #336, opened 2025-09-05
Simulated Annealing Implementation

blackhammer116:
## Description
This implementation of Simulated Annealing, Based on the C++ implementation,is a metaheuristic optimization algorithm designed to find the best instance (solution) in a search space by iteratively improving upon candidates. It mimics the physical process of annealing in metallurgy, where a material is heated and then slowly cooled to reach a stable state.

## How Has This Been Tested?
By Running the test cases written in the test file ```optimization/simulated-annealing-algo/test/simulated-annealing-test.metta```. the test checks if their are >= 4 instances and also if all of them are scored. The test for the actual score is commented out because the ```seed``` is not working properly as it should.

### DIFF
--- a/optimization/simulated-annealing-algo/simulated-annealing.metta
+++ b/optimization/simulated-annealing-algo/simulated-annealing.metta
@@ -40,12 +40,12 @@
 )
 
 
-;; Calculates instance energy using the composite score of that instance 
+;; Calculates instance energy using the composite score of that instance (
 ;;  (this is based on the C++ version) the composite score is the sum of
 ;;  the bScore and the penalized score.
 (: getEnergy (-> Cscore Number))
 (= (getEnergy $cScore)
-  (let ($pScore $bScore) ((getPenScore $cScore) (getScore $cScore)) (* -1 (+ $pScore $bScore)))
+  (let $pScore (getPenScore $cScore) (* -1 $pScore))
 )
 
 ;; Acceptance probability which is based on the energy of the current and new instance, and the temperature.
@@ -83,7 +83,6 @@
               ($deme (mkDeme (mkRep (mkKbMap (mkDscKbMp $idxMp) (mkDscMp $disc)) $tree) $instSet $id))
               ($table (mkTruthTableBScore $cpxCoeff $size $iTable))
               ($totalNNeighbors (estimateNeighborhood $d $disc))
-              ; (($alreadyXover $lastchance) (False False))
 
               (() (println! (Estimated neighbors: $totalNNeighbors)))
               (() (println! ""))
@@ -138,7 +137,6 @@
 
               ;; Check for acceptance
               ($isAccepted (> $actualAcceptanceProb $randomFloat))
-              ; ($isAccepted (or (< $newEnergy $energy) (< $randomFloat $acceptanceProb)))
               
               (() (println! (Is accepted: $isAccepted)))
               (() (println! (New instances: $newInstances)))


Yagth (commented):
Why are the `merge-demes-test` file renamed to `merge-demes-tests`
Same question for the `score-deme-test` file. The script won't pick them up otherwise

optimization/hillclimbing/test/hillclimbing-helpers-test.metta
```diff
@@ -101,13 +101,13 @@
                  (mkMultip 3) (mkDiscSpec 0) (mkDiscSpec 0) Nil)
                ))
             NilMMap))))))
-;; Test informationTheoreticBits with a knob 
-;; map containing two knobs with same multiplicity of 3
-!(assertEqual (informationTheoreticBits knobmap1) 6)
+; ;; Test informationTheoreticBits with a knob 
+; ;; map containing two knobs with same multiplicity of 3
```
Yagth, line 105: Why are these test cases commented out?

Yagth (commented):
Same question for the lsk-tests.metta and representation-tests.metta, their names should not include s in the end. Other wise the script won't pick them up during testing

optimization/simulated-annealing-algo/simulated-annealing.metta
```diff
@@ -0,0 +1,175 @@
...
+
+;; APPLY< helper for apply <
+(= (APPLY< $x $b))
+
+;; Calculates instance energy using the composite score of that instance 
+;;  (this is based on the C++ version) the composite score is the sum of
+;;  the bScore and the penalized score.
+(: getEnergy (-> Cscore Number))
+(= (getEnergy $cScore)
+  (let ($pScore $bScore) ((getPenScore $cScore) (getScore $cScore)) (* -1 (+ $pScore $bScore)))
```
Yagth, line 50: The way you calculate energy has differences with the C++ version of calculating energy, was that intentional?

optimization/simulated-annealing-algo/simulated-annealing.metta
```diff
@@ -0,0 +1,175 @@
...
+  (if (< $energy2 $energy1)
+    1.0
+    (math-exp (/ (- $energy1 $energy2) $Temp))
+  )
+)
+
+;; Actual acceptance probabitlity which is the product of the intensity and acceptance probability.
+;;  (This is based on the C++ implementation)
+(: actualAcceptanceProb (-> Number Number Number))
+(= (actualAcceptanceProb $acceptanceProb $intensity)
```
Yagth, line 65: This is a very simple function. Seems like it's not doing much so why not use the actual piece of code in place of the function call? That would make reading the code easier :)

optimization/simulated-annealing-algo/simulated-annealing.metta
```diff
@@ -0,0 +1,175 @@
...
+
+                          
+            )
+
+            (simulated-annealing $newDeme $table ;; $table
+                         $initCenter
+                         ($maxDist $minXoverNeighbors $bestPossibleScore) ;; $constParams
+                         ($newTemp $finalTemp $coolRate False $newIterCount) ;; $hyperParams
+                         ((and $xOver (not $hasImproved)) $lastChance $newCenter $newBestCScore $newBestScore $updatedNumInstances $newDistance (+ 1 $i))) ;; $params
+            ; (println! ("it worked"))
```
Yagth, line 172:
```suggestion
```

optimization/simulated-annealing-algo/simulated-annealing.metta
```diff
@@ -0,0 +1,175 @@
...
+              ($newBestCScore (if $isAccepted (Pair.second $newBestInstPair) $bestCScore))
+              ($newBestScore (if $isAccepted $newBestScore $bestScore))
+              ($updatedNumInstances (+ $currentNInstances $newInstances))
+              ($newTemp (- $Temp (* $coolRate $Temp)))
+              ($newIterCount (- $maxIter 1))
+
+                          
+            )
+
+            (simulated-annealing $newDeme $table ;; $table
```
Yagth, line 167:
```suggestion
            (simulated-annealing $newDeme $table
```

optimization/simulated-annealing-algo/test/simulated-annealing-test.metta
```diff
@@ -0,0 +1,91 @@
...
+
+
+!(assertEqual 
+    !(let*
+       (
+            (($instance (mkDeme $rep (mkSInstSet $instSet) $id) $state) (simulated-annealing deme table (mkInst (Cons 0 (Cons 0 Nil))) (20 0.01 0.1 True 5)))
+            ($instanceCount (List.length $instSet))
+            ($scoredElems (List.map isScored $instSet))
+            ; (() (println! "Instance count: " $instanceCount))
+            ; (() (println! "Scored elems: " $scoredElems))
```
Yagth, line 74:
```suggestion
```

Yagth (changes requested):

optimization/simulated-annealing-algo/simulated-annealing.metta
blackhammer116: its actually the same the difference is we use a score with the complexity included and i decided to use the cummilative sum of the scores, which would tell the actual score of the instance which will include the optimality

optimization/simulated-annealing-algo/simulated-annealing.metta
blackhammer116: yeah, i was trying to modularize most of the functions for ease but yeah i've used it directly as you suggested

blackhammer116:
Are there any comments left, I was gonna merge my local repo with the reduct so i could work on it?

optimization/simulated-annealing-algo/simulated-annealing.metta
blackhammer116: yeah the point was to combine both the complexity and the normal score. does that mean the complexity score doesn't matter because two instances can have the same score and the algo won't improve. that was the point of adding the complexity score

Yagth: I see. Then in that case, why not include the one with the complexity not the raw score. Then it will work fine. The composite score considers the complexity too so we can just use the penalized score :)

blackhammer116: so the penScore would do just fine right?

Yagth (approved):

PR #342, opened 2025-09-29
Implement confidence and Precison-Bscore scoring for feature Selection

Nama21yo:
## Description

This PR introduces two major updates to the feature selection scoring:

1. **Precision-BScore Implementation**

   * A  `precision_bscore` scorer has been added that was needed for feature selection.
   * It evaluates feature subsets based on **precision scoring** with support for penalties, activation thresholds.

2. **Confidence Scoring for Mutual Information (MI) Scorer**

   * A confidence weighting mechanism has been added to the MI scorer.
## Motivation and Context

* Other than `MutualInformation` the c++ moses also uses `precison-bscore` for feature set scoring.
* The **confidence factor** in MI scoring reduces overfitting by penalizing very small or very large feature subsets.

## How Has This Been Tested?

* Implemented unit tests for:

  * **Precision-BScore** result has been compared with the c++ test result and it is the same.
  *  for the rest updates too.

## Types of changes

* [x] New feature (non-breaking change which adds functionality)
* [ ] Bug fix (non-breaking change which fixes an issue)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

* [x] My code follows the code style of this project.
* [x] I have added tests to cover my changes.
* [x] All new and existing tests passed.


scoring/precision-bscore.metta
```diff
@@ -0,0 +1,351 @@
...
+          ;; Extract corresponding labels
+          ($filteredLabels (map-atom $exprIndices $each (List.at $each $labels)))
+          ($filteredExpr (List.listToExpr $filteredRows))
+          )
+     ($filteredExpr $filteredLabels)
+   )
+)
+
+;; Main pre_scorer implementation 
+;; This is the top-level function called by hill climbing for feature selection
```
Yagth, line 303: As far as I know. This isn't the only scorer being called by the hill climber. It actually uses `mi` by default doesn't it? Unless the user specifically specifies the socrer to be `pre_scorer`?

Yagth: Regardless, this is still a useful addition to the current scorer we have

Yagth (approved):

PR #344, opened 2025-09-29
Reduct refactoring and integration

blackhammer116:
## Description
This reduct version is taken from previously implemented reduct-code inside the `feat/ReductIntegration` which was partially completed. Inside this branch the main issue was that it wasn't adhearing to the rules of reduction according to _Elements of an Expert System for Determining the Satisfiability of General Boolean Expressions" by Craig Stewart Holman (1990)_ . Moreover the implemented `rte` functions weren't reducing properly that would allow moses to run at least 1 generation

## How Has This Been Tested?
- Except the representation all functions have their own testcases which can be found under `tests/` folder inside their respective parent folders.
- Since the representation is running with semi-random and random algorithm their exact results can't be test, however all its respective algorithms and helper functions have been thoroughly tested which can be found under `representation/tests/` and `feature-selection/tests/`

## Types of changes
- Inside the reduce function, after `propagateNot` and `gatherJunctors` function have ran `AND` nodes will be added to the expression as a parent node if it doesn't contain that already.
- After an inconsistant node has been deleted by `deleteInconsistantHandle` it will get replaced by an empty `AND` node which will be caught and removed by `andCut`.
- If an empty `AND` node is found inside an `OR` expression the entire `OR` expression will be removed by  `zero-constraint-subsumption`.
- Also `zero-constraint-subsumption` handles the removal of an empty `AND` node at the end if its not the parent expression.

## Checklist:


utilities/ordered-set.metta
```diff
@@ -142,12 +142,15 @@
 ;; Returns: (OS Number) → Union of both sets
 ;; Example: (OS.union (ConsOS 0 (ConsOS 1 NilOS)) (ConsOS 2 NilOS)) → (ConsOS 0 (ConsOS 1 (ConsOS 2 NilOS)))
 (: OS.union (-> (OS Number) (OS Number) (OS Number)))
-(: OS.union (-> (OS Number) (OS Number) (OS Number)))
-(= (OS.union NilOS $set2) $set2)
-(= (OS.union (ConsOS $head $tail) $set2)
-   (chain (OS.insert genericCompare $head $set2) $newSet2
-     (OS.union $tail $newSet2)
-   )
+(= (OS.union $set1 $set2)
```
Yagth, line 145: I can't stop but wonder, since this change was previously implemented using normal function definitions instead of a `unify` function, it was a result of incorrect merge conflict handling. The `unify` isn't stable and causes errors. If the change wasn't intentional can you make sure the previous change is used instead?

utilities/ordered-set.metta
```diff
@@ -158,13 +161,16 @@
...
 (: OS.difference (-> (OS Number) (OS Number) (OS Number) (OS Number)))
-(= (OS.difference NilOS $set2 $result) $result)
-(= (OS.difference (ConsOS $head $tail) $set2 $result)
-   (if (OS.contains $set2 $head)
-       (OS.difference $tail $set2 $result)
-       (chain (OS.insert genericCompare $head $result) $newResult
-         (OS.difference $tail $set2 $newResult)
-       )
+(= (OS.difference $set1 $set2 $result)
+   (unify $set1 NilOS
```
Yagth, line 165: The same goes for this one as well. You can review your own PR before sending it over to make sure you can catch this kind of differences between your change and the one on the `ref/mettalog` branch.

utilities/map.metta
```diff
@@ -91,14 +91,14 @@
 ;; Get the index of a key value pair given a key
 (: Map.find (-> (Map ($k $v)) $k Number))
 (= (Map.find NilMap $k) -1)
-(= (Map.find (ConsMap ($curKey $curVal) $xs) $k')
-   (if (== $curKey $k')
+(= (Map.find (ConsMap ($k $v) $xs) $k')
+   (if (== $k $k')
        0
-       (chain (Map.find $xs $k') $targetIdx
+       (chain (Map.find $xs $k'' ) $targetIdx
```
Yagth, line 97: I don't understand why the `$k'` is changed into `$k''` there's no such variable definition.


```suggestion
       (chain (Map.find $xs $k') $targetIdx
```

utilities/tests/lru-cache-test.metta
```diff
@@ -1,99 +0,0 @@
-!(register-module! ../../../metta-moses)
```
Yagth, line 1: This test file should never have been deleted.

utilities/tests/list-methods-test.metta
```diff
@@ -219,10 +219,4 @@
 !(assertEqual (List.generate 3 5) (Cons 5 (Cons 5 (Cons 5 Nil))))  
 !(assertEqual (List.generate 5 3) (Cons 3 (Cons 3 (Cons 3 (Cons 3 (Cons 3 Nil))))))
 !(assertEqual (List.generate 0 3) Nil)
-!(assertEqual (List.generate 3 0) (Cons 0 (Cons 0 (Cons 0 Nil))))
-
-;; Normal removal from head
```
Yagth, line 224: Why are this test cases removed?

utilities/tests/general-helper-functions-test.metta
```diff
@@ -458,120 +450,4 @@ NilMMap discSpec<=)
 ;; Test cases -- concatT
 ! (assertEqual (concatT (1 2 3) (4 5 6)) (1 2 3 4 5 6))
 ! (assertEqual (concatT (1 2 3) ()) (1 2 3))
-! (assertEqual (concatT () (1 2 3)) (1 2 3))
-
```
Yagth, line 462: Why did you remove these test cases?

scoring/cacheSpace.metta
```diff
@@ -1,2 +0,0 @@
-;; 
-! (bind! &cacheSpace (new-space))
```
Yagth, line 2: There are still files that are using this import. Can you place it back?

representation/lsk.metta
```diff
@@ -14,22 +14,24 @@
...
-    ($reducedSubtree $negatedSubtree)  ;;;after the reduct integration, the reduction step will be applied here 
+    ; (() (println! (NegatedSubtree : $negatedSubtree)))
+    ($reducedSubtree (cleanTree $negatedSubtree))  ;;;after the reduct integration, the reduction step will be applied here 
+    ; (() (println! (ReducedSubtree: $reducedSubtree))) 
     ($childrenOfTarget (getChildrenById $tree (mkNodeId $target)))
    )
    (if (List.contains $subtree $childrenOfTarget) 
        (let $idOfSubtree (getSubtreeId $tree (mkNodeId $target) $subtree 0)
-           ($tree (mkLSK (mkDiscKnob (mkKnob (updateID $idOfSubtree)) (mkMultip 3) (mkDiscSpec 1) (mkDiscSpec 1) Nil)))
+           ($tree (mkLSK (mkDiscKnob (mkKnob $idOfSubtree) (mkMultip 3) (mkDiscSpec 1) (mkDiscSpec 1) Nil)))
```
Yagth, line 24: Any particular reason for removing the `updateID` function call? It was handling a weird edge case when it was implemented?

Yagth: I am pretty sure the leading zeros are a result of the removal of this function too.

representation/tests/create-representation-test.metta
```diff
@@ -65,5 +77,7 @@
...
...SK (mkDiscKnob (mkKnob (mkNodeId (4))) (mkMultip 3) (mkDiscSpec 0) (mkDiscSpec 0) Nil))) (ConsMMap ((mkDiscSpec 3) (mkLSK (mkDiscKnob (mkKnob (mkNodeId (3))) (mkMultip 3) (mkDiscSpec 0) (mkDiscSpec 0) Nil))) (ConsMMap ((mkDiscSpec 3) (mkLSK (mkDiscKnob (mkKnob (mkNodeId (0 1))) (mkMultip 3) (mkDiscSpec 1) (mkDiscSpec 1) Nil))) (ConsMMap ((mkDiscSpec 3) (mkLSK (mkDiscKnob (mkKnob (mkNodeId (2))) (mkMultip 3) (mkDiscSpec 0) (mkDiscSpec 0) Nil))) NilMMap)))))))))) (mkTree (mkNode AND) (Cons (mkTree (mkNode OR) (Cons (mkTree (mkNode A) Nil) (Cons (mkNullVex (Cons (mkTree (mkNode B) Nil) Nil)) (Cons (mkNullVex (Cons (mkTree (mkNode AND) (Cons (mkTree (mkNode B) Nil) (Cons (mkTree (mkNode A) Nil) Nil))) Nil)) (Cons (mkNullVex (Cons (mkTree (mkNode AND) (Cons (mkTree (mkNode NOT) (Cons (mkTree (mkNode B) Nil) Nil)) (Cons (mkTree (mkNode A) Nil) Nil))) Nil)) Nil))))) (Cons (mkNullVex (Cons (mkTree (mkNode B) Nil) Nil)) (Cons (mkNullVex (Cons (mkTree (mkNode OR) (Cons (mkTree (mkNode B) Nil) (Cons (mkTree (mkNode A) Nil) Nil))) Nil)) (Cons (mkNullVex (Cons (mkTree (mkNode OR) (Cons (mkTree (mkNode NOT) (Cons (mkTree (mkNode B) Nil) Nil)) (Cons (mkTree (mkNode A) Nil) Nil))) Nil)) Nil)))))))
```
Yagth, line 80: Here is an example of a leading zero that might be as a result of the removal of the `updateID` function.

Yagth (changes requested):

PR #377, opened 2025-11-26
PeTTa LRU cache, feature selection, similarity and precision scorers, cross-top-one

K1rubel:
## Description
Refactoring existing code to work with `PeTTa`. There are some differences to reconcile with this `MeTTaLog` refactored code base. In addition to the type system, which is not working, there are some behavioral differences noticed and corrected for using work arounds. 

- `unify` seems to have issue when nested -- so `=` was used instead for unification with `if` like `(if (= $pattern1 $pattern2)  $then $else)`
- `union-atom` used to return unique set of values in `MeTTaLog` and it was used in places like `powerSet` computation. But, `PeTTa's` union-atom implementation doesn't, which is the expected behaviour. A custom implementation that maintains unique entries only was implemented. 

## Types of changes

## Checklist:


Yagth:
I am going to merge this in order not to block the other devs. But address the conversations in a separate PR  :)

PR #395, opened 2025-12-17
Fix getCScore else evaluation and prevent unintended atom-space update and empty return

Cheralia:
## Description
In Petta, function arguments are evaluated during parameter assignment at call time. As a result, even when a parameter is not used in the function body, its value is still executed simply because it appears in the function definition. 
This issue affects the **getCScore** function, which calls unify from **utilities/general-helpers.metta**. 
The expected behavior is:
- If the pattern does not exist in the atom-space, compute the score up to updatePenalizedScore and then add the pattern to the space.
- If the pattern already exists, simply return the stored score without recomputation.

However, because the values passed to unify eagerly evaluate both the then and else alternatives, the else branch is executed regardless of whether the pattern exists in the atom-space. This execution happens in the function definition stage rather than in the function body, leading to multiple unintended results being added to the atom-space.

This behavior causes issues when getCScore is executed more than once in a single run.

Additionally, unify was returning Empty when the pattern did not exist, even after executing the else branch. This caused downstream functions (directly or indirectly dependent on getCScore) to halt or fail when attempting to use the returned value.

## Motivation and Context
While testing the univariate optimization on PR #391 , the last test case exhibited an issue where execution entered an endless loop. Although the test output repeatedly reported true, the evaluation continued indefinitely.

Further analysis revealed that the root cause was related to the getCScore function. When getCScore is called two or more times within a single execution, the issue becomes reproducible. This behavior was also observed in the score-deme test case, where getCScore is invoked multiple times.

The problem stems from unintended memory updates caused by eager evaluation, as well as Empty return values that interrupt control flow. These issues lead to **repeated evaluations**, **unnecessary memory growth**, and **execution halts** in dependent logic. Addressing time and memory optimization, preventing invalid Empty returns, and ensuring stable execution when calling getCScore and evaluating its values are essential for correct termination and reliable behavior of functions that depend on it.

## How Has This Been Tested?
- Verified that all getCScore test cases pass consistently across multiple executions.
- Confirmed that all univariate optimization test cases pass repeatedly without delays or the indefinite execution behavior observed previously.
- Verified that all score-deme test cases pass consistently across repeated runs.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/deme/tests/score-deme-test.metta
+++ b/deme/tests/score-deme-test.metta
@@ -123,13 +123,13 @@
 
 ; The these test cases are commented out because they behave non deterministically when running all the 3 representation together 
 ;; but runs correctly when running them separately !!
-; !(assertEqual
-;    (transform
-;    (mkSInstSet (Cons (mkSInst (mkPair (mkInst (Cons 0 (Cons 1 (Cons 2 Nil)))) -1.0000000000000001e308)) (Cons (mkSInst (mkPair (mkInst (Cons 0 (Cons 1 (Cons 1 Nil)))) -1.0000000000000001e308)) Nil)))
-;    (mkRep (knobMapObj) (tree1))
-;    (ttable2)
-;    1)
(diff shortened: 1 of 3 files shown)


Yagth:
@Cheralia I don't understand why, according to your description, the error would arise. But the issue was indeed because of a duplicate definition of the `unify` function. You need to remove one of them. Even though your current implementation works just fine, the underlying issue is still not resolved. Which is the duplicate function definition. And it will also affect anywhere else where it has been used.

Yagth (approved):

PR #445, opened 2026-05-08
feat(crossover): implement legacy MOSES crossover operators

Yadelew-zemene:
## Description

This PR introduces multi-parent crossover capability from legacy MOSES into the MeTTa-based crossover system.
It restores and adapts crossTopTwo and crossTopThree to work with the current architecture, enabling stable multi-parent recombination in hill-climbing optimization.

## Main Motivation
   -  Improve diversity and exploration in hill-climbing search
   
##  What this enables :
          -    Supports 2-parent and 3-parent crossover in the current system
          -    Restores missing legacy MOSES exploration behavior
          -    Improves search diversity during hill-climbing optimizations
         
##  Changes
         -      Built crossTopTwo (pairwise crossover)
         -      Built crossTopThree (triple crossover)
         -      Added helper utilities for instance handling and merging
         -      Integrated operators into existing crossover pipeline

## Testing
         -     Unit tests for crossTopTwo, crossTopThree, and helpers
         -     Structural validation of crossover flow 
         -     End-to-end pipeline test on sample demes

All tests passed successfully.


optimization/hillclimbing/hill-climbing-helpers.metta
```diff
@@ -120,11 +120,26 @@
...
+;    ;; INFO: The 120 is crossover_pop_size parameter that should come from hillClimbing
+;    ;;;..this line is ...Prevous implementation ......  ((crossTopOne $deme (min $nNewInstances 120) $sampleStart $sampleSize $centerInst) (if (< (- $sampleSize 1) $nNewInstances) (- $sampleSize 1) $nNewInstances))
+(let*
+    (
+      ($nToMake (min $nNewInstances 120))
+      ((mkDeme $rep (mkSInstSet $instSet) $id) $deme)
+      
+      ($newInst1 (crossTopOne $instSet (+ (/ $nToMake 3) (% $nToMake 3)) $sampleStart $sampleSize $centerInst))
+      ($newInst2 (crossTopTwo $instSet (/ $nToMake 3) $sampleStart $sampleSize $centerInst))
+      ($newInst3 (crossTopThree $instSet $n3 $sampleStart $sampleSize $centerInst))
```
Yagth: Where `$n3` coming from here? Is that a bug?

optimization/hillclimbing/test/hillclimbing-helpers-test.metta
```diff
@@ -248,4 +252,25 @@
...
+)
+!(let
+    ($d1 (crossTopOne (deme1) 3 0 4 (mkInst (cons 0 ()))))
+    ($d2 (crossTopTwo $d1 3 0 4 (mkInst (cons 0 ()))))
+    (assertTrue (not (== $d2 ()))))
```
Yagth: Please do not use an undefined function like this. Use the correct assert function.

optimization/hillclimbing/hill-climbing-helpers.metta
Yadelew-zemene: corrected

Yagth:
The test  case that used to finish in 2mins took an hour. Canceling it for now.

optimization/hillclimbing/test/hillclimbing-helpers-test.metta
```diff
@@ -246,6 +251,26 @@
...
+   (mkTree (mkNode PRIORITIZED-OR) ((mkTree (mkNode playwin) ()) (mkTree (mkNode playblock) ()))))
+  (mkSInstSet ((mkSInst (mkPair (mkInst (0 1)) (mkCscore -3.0 1 0.2857142857142857 0.0 -3.2857142857142856)))
+                (mkSInst (mkPair (mkInst (1 1)) (mkCscore -3.0 2 0.5714285714285714 0.0 -3.571428571428571)))
+                (mkSInst (mkPair (mkInst (0 1)) (mkCscore -3.0 1 0.2857142857142857 0.0 -3.2857142857142856)))
+                (mkSInst (mkPair (mkInst (1 0)) (mkCscore -3.0 1 0.2857142857142857 0.0 -3.2857142857142856)))))
+  (mkDemeId "1"))
+ (true false false (mkInst (1 1)) 3 1 (mkCscore -3.0 1 0.2857142857142857 0.0 -3.2857142857142856) -3.2857142857142856 4 2 3)))
+
+
+!(assertEqual (hillClimbing (strategyDeme) (mkCscore 3.0 5 1.43 0.0 3) (mkInst (cons 1 (cons 1 ()))) (mkParams applyStrategyBasedScore (3 random-player 3.5)))
```
Yagth:
```suggestion
!(assertEqual (hillClimbing (strategyDeme) (mkCscore 3.0 5 1.43 0.0 3) (mkInst (cons 1 (cons 1 ()))) (mkParams applyStrategyBasedScore (3 3.5)))
```

Yagth:
@Yadelew-zemene can you resolve the conflicts? This is ready to be merged.

Yadelew-zemene:
> @Yadelew-zemene can you resolve the conflicts? This is ready to be merged.

Sure. I will see that.

Yadelew-zemene:
> > @Yadelew-zemene can you resolve the conflicts? This is ready to be merged.
> 
> Sure. I will see that.

Check It and let me know if any more conflict.

Yagth:
Hey @Yadelew-zemene I think the way you handled merge conflict is causing one of the tests to fail. Can you check?

Yagth (changes requested):
Hey Yadelew. I tried running your code locally and found the following issues with the expand-deme-test file, cause for it's failure. The first thing is that you have two `cross-topone-helper` file imports and a missing `crossover-helpers` file which I guess is caused unintentionally because of the merge conflict resolution. 

The second thing I noticed was the use of the `getInst` function (which was copied into the `crossover-helpers` file is a duplicate function that also lives in the `instance.metta` that file is imported after the `crossover-helpers files` file. Imports work weird in PeTTa so the function in that file is sometimes running on other times it isn't, causing non deterministic result when the copied function is left untouched but breaking the code when the copied function gets commented out. I suggest you rename the copied function and all of it's usages for the sake of compatibility for now and we will find a better fix later. 

The hillclimbing test case is failing for the strategy knobs. Can you simply check the test case and see what the cause of the difference in results are? If you can manage to find out why let me know and we can figure out a way to correct it, otherwise if all seems to be corrected and the new result is actually the intended result, then we will update it. My guess is that this is also somehow messed up after the merge conflict resolution.

After you are done with this updates, please make sure to also test it on the parity3 problem with 1000 number of evals, single deme and no feature selection algorithm with seed 2. If you are having trouble running this test, let me know and I can assist. However we need to verify that the current implementation doesn't cause a regression in result quality.

Yadelew-zemene:
> Hey Yadelew. I tried running your code locally and found the following issues with the expand-deme-test file, cause for it's failure. The first thing is that you have two `cross-topone-helper` file imports and a missing `crossover-helpers` file which I guess is caused unintentionally because of the merge conflict resolution

Hi,@Yagth 
Sorry for late.

Thank you for pointing these issues. To avoid the import issue for  crossover operators dependent test files  I decided to host  all cross crossover helpers in cross-top-one , ran tests and  found that the issue in expand-deme-test file is the way merge3demes produce results. so I need to adjust it so that it produce the expected structure even though operators produce same deme structure and I will make an update for these changes.

Yadelew-zemene:
> The second thing I noticed was the use of the `getInst` function (which was copied into the `crossover-helpers` file is a duplicate function that also lives in the `instance.metta` that file is imported after the `crossover-helpers files` file. Imports work weird in PeTTa so the function in that file is sometimes running on other times it isn't, causing non deterministic result when the copied function is left untouched but breaking the code when the copied function gets commented out. I suggest you rename the copied function and all of it's usages for the sake of compatibility for now and we will find a better fix later.


For this  issue  I think we need not to define getInst at all.  I have just removed it and tested , and found using the already defined getInst results in similar way , and outputs are deterministic. so we can remove the new one.

Yadelew-zemene:
> The hillclimbing test case is failing for the strategy knobs. Can you simply check the test case and see what the cause of the difference in results are? If you can manage to find out why let me know and we can figure out a way to correct it, otherwise if all seems to be corrected and the new result is actually the intended result, then we will update it. My guess is that this is also somehow messed up after the merge conflict resolution.

The reason  hillclimbing test case is failing for the strategy knobs is becuase of  i adjusted the transformation function calling   in the hillclimbing-helpers implementation as  

(case $iTable
                                            (
                                              ((mkITable $rows $labels) ($transformFunction $updatedInstSet $rep $id (mkITable $rows $labels) $rowOCoeff))
                                              ($ComplexityRatio ($transformFunction $updatedInstSet $rep $id ($rowOCoeff) $ComplexityRatio))
                                            ))))


 Previously $id  arg was not passed.But in the updated  tranform method defined  in score- deme.metta,it takes the $id parameter as 
      (= (transform (mkSInstSet $scoredInstList) $rep $id $scoringContext $complexityRatio)
   (let* (
          ($instList (List.map removeInstScore $scoredInstList))
          ($scored (case $scoringContext.
          
     I prefer if you explian these .

Yagth:
> Hi,@Yagth Sorry for late.
> 
> Ran tests and found that the issue in expand-deme-test file is the way merge3demes produce results. so I need to adjust it so that it produce the expected structure even though operators produce same deme structure and I will make an update for these changes.

Changing the name of the duplicate function `getInst` living in both instance.metta and one of the new crossover helpers was the second cause I noticed yadelew. Change the name of the second set of the function seem to have fixed the issues in the expand-deme-test file on my side.

Yagth:
> > The hillclimbing test case is failing for the strategy knobs. Can you simply check the test case and see what the cause of the difference in results are? If you can manage to find out why let me know and we can figure out a way to correct it, otherwise if all seems to be corrected and the new result is actually the intended result, then we will update it. My guess is that this is also somehow messed up after the merge conflict resolution.
> 
> The reason hillclimbing test case is failing for the strategy knobs is becuase of i adjusted the transformation function calling in the hillclimbing-helpers implementation as
> 
> (case $iTable ( ((mkITable $rows $labels) ($transformFunction $updatedInstSet $rep $id (mkITable $rows $labels) $rowOCoeff)) ($ComplexityRatio ($transformFunction $updatedInstSet $rep $id ($rowOCoeff) $ComplexityRatio)) ))))
> 
> Previously $id arg was not passed.But in the updated tranform method defined in score- deme.metta,it takes the $id parameter as (= (transform (mkSInstSet $scoredInstList) $rep $id $scoringContext $complexityRatio) (let* ( ($instList (List.map removeInstScore $scoredInstList)) ($scored (case $scoringContext.
> 
> ```
>  I prefer if you explian these.
> ```

Well that must have been a regression introduced after solving the merge conflict yadelew, because the correct behavior is the transform function with the `$id` therefore if you have to update your test case to match that fine. However, the original cause of the updated result I saw is the difference in the instances output. Originally they were all duplicates `(1 1)` instance all three of them being returned by the hill climber for the strategy knob. However, after the merge the instances got more diverse. I don't hate that is the case (in fact it is appreciated), I am just trying to understand why that came to be the case.

Yadelew-zemene:
> > > The hillclimbing test case is failing for the strategy knobs. Can you simply check the test case and see what the cause of the difference in results are? If you can manage to find out why let me know and we can figure out a way to correct it, otherwise if all seems to be corrected and the new result is actually the intended result, then we will update it. My guess is that this is also somehow messed up after the merge conflict resolution.
> > 
> > 
> > The reason hillclimbing test case is failing for the strategy knobs is becuase of i adjusted the transformation function calling in the hillclimbing-helpers implementation as
> > (case $iTable ( ((mkITable $rows $labels) ($transformFunction $updatedInstSet $rep $id (mkITable $rows $labels) $rowOCoeff)) ($ComplexityRatio ($transformFunction $updatedInstSet $rep $id ($rowOCoeff) $ComplexityRatio)) ))))
> > Previously $id arg was not passed.But in the updated tranform method defined in score- deme.metta,it takes the $id parameter as (= (transform (mkSInstSet $scoredInstList) $rep $id $scoringContext $complexityRatio) (let* ( ($instList (List.map removeInstScore $scoredInstList)) ($scored (case $scoringContext.
> > ```
> >  I prefer if you explian these.
> > ```
> 
> Well that must have been a regression introduced after solving the merge conflict yadelew, because the correct behavior is the transform function with the `$id` therefore if you have to update your test case to match that fine. However, the original cause of the updated result I saw is the difference in the instances output. Originally they were all duplicates `(1 1)` instance all three of them being returned by the hill climber for the strategy knob. However, after the merge the instances got more diverse. I don't hate that is the case (in fact it is appreciated), I am just trying to understand why that came to be the case.

Right. Updating the test accourdigly is just adjusting . But what i see is that the hill climber  test for the strategy knob is not reduced at all on dev if you notice . Due to this I update the tranform function calling and it made the test reduced which results in output difference. 
I  would appriciate  if you could  test   "hill climber  test for the strategy knob"   on dev. Before running it,make sure to comment the two tests under  !(hillclimbing test ----------), as they hide it.



> > Hi,@Yagth Sorry for late.
> > Ran tests and found that the issue in expand-deme-test file is the way merge3demes produce results. so I need to adjust it so that it produce the expected structure even though operators produce same deme structure and I will make an update for these changes.
> 
> Changing the name of the duplicate function `getInst` living in both instance.metta and one of the new crossover helpers was the second cause I noticed yadelew. Change the name of the second set of the function seem to have fixed the issues in the expand-deme-test file on my side.

Yadelew-zemene:
Closing this PR in favor of implementation on a new branch. The work has been reorganized and Further updates and review will continue there.

PR #450, opened 2026-05-26
ref: rewrite candidate reconstruction bottom-up (~2x faster)

Leul-Negash:
## Description
Rewrites instance-to-tree candidate reconstruction in `representation/representation.metta` from a top-down builder to a bottom-up one.

The old `getCandidateRec` threaded a growing candidate tree and inserted each node with `appendChild` (which does `getNodeById` + `replaceNodeById` on the candidate) once per node. The new builder (`buildNodeBU` / `buildChildrenBU` / `buildCopyBU` / `buildKnobBU`) assembles each subtree bottom-up and returns it, so a parent gets its children already built and never walks or rebuilds a growing candidate. Knob semantics (present, NOT-wrap for d=2, absent, and null-vertex unwrapping) are reproduced exactly. The dead `applyGetCandidateRec` helper is removed; the `appendTo*` helpers stay (still exercised directly by `lsk-test.metta`).

Also adds `docs/performance-bottlenecks.md`, the KR1 profiling write-up that
identified this hotspot and documents the remaining bottlenecks.
Review order: `representation/representation.metta` first, then the doc.

## Motivation and Context
Objective 2, KR1 (identify and document the main performance bottlenecks) and KR2 (optimize the critical paths). Phase profiling showed `getCandidate` was ~98% of a `majority3` run (73.4s of ~75s; all scoring was 1.1s). The cost was the candidate-side `appendChild`/`replaceNodeById` rebuilds, which grow with the candidate. Reconstruction was the single largest opportunity in the pipeline.

## How Has This Been Tested?
- PeTTa v1.0.0 via `run.sh`.
- `representation/tests/representation-test.metta`: `getCandidateRec` across 6
  instances (0 0 0 / 1 1 1 / 2 2 2 / 2 1 0 / 1 0 2 / 2 0 1) and `getCandidate` all
  pass, output unchanged.
- Full suite `python3 scripts/run-tests.py`: 76/76 pass.
- `majority3` benchmark (seed 0, maxGen 3, 1 deme): ~75s before, ~39.5s after
  (~48% faster), `FinalResult` byte-identical between the two.

## Types of changes
- [x] Performance refactor (non-breaking; output is byte-identical)

## Checklist:
- [ ] I have added tests to cover my changes. (No new tests: behavior is
      byte-identical and already covered by existing representation and suite tests.)

### DIFF
--- a/docs/performance-bottlenecks.md
+++ b/docs/performance-bottlenecks.md
@@ -1,0 +1,306 @@
+# MOSES (MeTTa) performance bottlenecks
+
+Quarter plan reference: Objective 2 (Improve Performance and Runtime Efficiency),
+KR1: "Identify and document the main performance bottlenecks."
+
+This document records the first systematic profiling pass over the MeTTa MOSES
+pipeline and the bottlenecks it found. Each entry gives where the cost is, why it
+is slow, its measured or estimated impact, and a recommended fix mapped to a KR.
+It is meant to direct KR2 (optimize the critical execution paths) and KR3 (use
+more native PeTTa/Prolog).
+
+## Summary
+
+On PeTTa v1.0.0, a small Boolean problem (majority3: 8-row truth table, 3
+generations, 1 deme) takes about 75s wall and 700-880 MB peak RSS. CPU sits at
+99%, so the engine is single-threaded and scaling depends entirely on per-core
+efficiency.
+
+The dominant cost is candidate reconstruction (getCandidate), about 98% of runtime
+on majority3. All scoring together is about 1.5%. This was not obvious from the
+interpreter's own diagnostics: the loudest signal (the "Not specialized" warnings)
+pointed elsewhere, and the real cost only showed up under phase timers.
+
+Two findings worth carrying into the optimization work:
+
+- Warning count is not a proxy for time. The single largest "Not specialized"
+  source (expToMMap, 181 of 206 warnings) is about 2-3% of runtime, while
+  reconstruction emits no warnings and is about 98%.
+- Prioritize by measured wall time. Several plausible-looking changes moved wall
+  time by 3% or less; the large win came only after timers located the real
+  hotspot.
+
+## Methodology
+
+Measured on the target interpreter (PeTTa v1.0.0 via run.sh), using four signals:
+
+1. Wall-clock and resource usage from `/usr/bin/time -v` on a full runDemoProblem
+   run.
+2. Interpreter specialization warnings. PeTTa prints `Not specialized
+   <pred>/<arity>` when it falls back from compiled execution to the
+   meta-interpreter. Counting these by function points at code the runtime cannot
+   compile.
+3. Static complexity analysis, tracing the call graph from runMoses down through
+   deme expansion, sampling, scoring, reduction, and merge.
+4. Phase timers: `(py-call (time.time))` brackets around the expand/optimize/merge
+   phases of runMoses, and around getCandidate vs getCscore per candidate. These
+   located the dominant cost after the first three signals pointed at the wrong
+   place.
+
+This is wall-clock plus warning counts plus call-graph analysis, not a line-level
+statistical profiler (PeTTa does not expose per-predicate timing). The numbers are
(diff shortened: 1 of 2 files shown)


Yagth:
@LNAMTY  This looks like a good implementation, we have an even more optimized version which only takes around 5 seconds for the same problem configuration implemented at the branch `ref/RepresentationBuilding`. This branches works at this commit of PeTTa which is more latest than v1.0.0. `fb14a33` test it out on your side, compare the two and let me know your results. If it's better than what we have know, I will merge it. If not, we still have some issues to address with the new implementation (it is not stable yet) so we will keep it around as a fallback. Thanks for your contribution.

Leul-Negash:
> @LNAMTY This looks like a good implementation, we have an even more optimized version which only takes around 5 seconds for the same problem configuration implemented at the branch `ref/RepresentationBuilding`. This branches works at this commit of PeTTa which is more latest than v1.0.0. `fb14a33` test it out on your side, compare the two and let me know your results. If it's better than what we have know, I will merge it. If not, we still have some issues to address with the new implementation (it is not stable yet) so we will keep it around as a fallback. Thanks for your contribution.

Thanks for the pointer. I tried the comparison on my machine but I want to flag a couple of things first.
I couldn't find the PeTTa commit fb14a33 you mentioned. It's not on trueagi-io/PeTTa (the patham9/PeTTa URL in setup.sh redirects there), and I went through the public forks too. So I ran both branches on PeTTa v1.0.3 (ebdb511), which is the latest tag and is newer than v1.0.0. If fb14a33 is somewhere private, can you point me at the right place? Some of the speed gap might come from there.

Same machine, no other load, runDemoProblem 0 10 1 None <problem>. On majority3, ref/bottom-up-reconstruction took 172s and ref/RepresentationBuilding took 134s. On mux6 the RepresentationBuilding run finished in 311s; I killed the bottom-up run at 632s because it was still going. So RepresentationBuilding is faster on both, around 1.3x on majority3 and at least 2x on mux6. I didn't get anywhere close to the ~5s figure for majority3 on v1.0.3, which is the other reason I'm asking about the PeTTa commit.

On stability though, I hit a clear hang. representation/tests/build-knobs-test.metta ran for over 11 minutes on ref/RepresentationBuilding without printing a single line before I killed it. The same test finishes in seconds on ref/bottom-up-reconstruction and on dev, so this isn't a slow path, it's a hang. That goes with what you said about it not being stable yet.
Let me know what you think.

Yagth:
@LNAMTY The 5second run was for majority3 with 3 max generations, seed 0 and 1 deme just like the comparison you used for the PR. 

The PeTTa I tried to reference was just to avoid a potential compatibility issue, PeTTa doesn't have big difference in performance after 1.0,0.

As the result of your comparison, I think it is best to stick with our current version for now and try to address some of the issues I mentioned earlier. However, if doing that proves not feasible, I will reopen this PR.

Leul-Negash:
> @LNAMTY The 5second run was for majority3 with 3 max generations, seed 0 and 1 deme just like the comparison you used for the PR.
> 
> The PeTTa I tried to reference was just to avoid a potential compatibility issue, PeTTa doesn't have big difference in performance after 1.0,0.
> 
> As the result of your comparison, I think it is best to stick with our current version for now and try to address some of the issues I mentioned earlier. However, if doing that proves not feasible, I will reopen this PR.

I agree.  Is there any specific thing I can work on to address the issues as u said? If u have an idea on the things which are creating the problem.

Yagth:
> > @LNAMTY The 5second run was for majority3 with 3 max generations, seed 0 and 1 deme just like the comparison you used for the PR.
> > The PeTTa I tried to reference was just to avoid a potential compatibility issue, PeTTa doesn't have big difference in performance after 1.0,0.
> > As the result of your comparison, I think it is best to stick with our current version for now and try to address some of the issues I mentioned earlier. However, if doing that proves not feasible, I will reopen this PR.
> 
> I agree. Is there any specific thing I can work on to address the issues as u said? If u have an idea on the things which are creating the problem.

I will let you know if I need help with anything specific thanks :)

PR #454, opened 2026-05-29
fix: corrected the state update of the best score in the hillclimbing 

Liya-F:
There was an issue with the hill climbing optimization implementation where $bestScore and $bestSscore were updated unconditionally every iteration, regardless of whether a meaningful improvement had occurred. The $hasImproved check, which determines whether the new score exceeds the previous best by a threshold of 0.5, was only applied to the center instance update, not to the score state update. This created a misalignment between the center instance and its corresponding best score: the score baseline would drift to reflect a candidate that was never actually adopted as the new center, causing subsequent $hasImproved checks to compare against an inconsistent reference point.

This fix ensures that $bestScore and $bestSscore are only updated when $hasImproved is True, keeping them consistent with the current center instance at every iteration. This also improves the performance of the optimization algorithm by preventing premature score inflation from raising the improvement threshold, which previously caused genuinely better candidates to be incorrectly rejected and the search to terminate earlier than it should have.

## How Has This Been Tested?
It has been tested with the latest version of petta. 

## Types of changes

## Checklist:

### DIFF
--- a/optimization/hillclimbing/hill-climbing-helpers.metta
+++ b/optimization/hillclimbing/hill-climbing-helpers.metta
@@ -325,6 +325,8 @@
         ;; ($_ (println! ""))
 
         ($nextCenterInst (if $hasImproved (Pair.first $newBestInstPair) $centerInst))
+        ($nextBestScore (if $hasImproved $newBestScore $bestScore))
+        ($nextBestSscore (if $hasImproved (Pair.second $newBestInstPair) $bestSscore))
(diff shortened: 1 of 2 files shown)


Yagth (commented):
@Liya-F did you confirm that this fixed the drifting center? If so can you also add a test case to add catch that issue incase it happens in the future? A good test case would be one that will fail without this update but passes with this. It can be an exact check on the center to see if it has changed or not after `n` number of iterations for example.

Yagth (approved):

PR #468, opened 2026-06-20
Ref/Storing Demes inside Atomspace implementaion

Liya-F:
Refactors deme storage from the mkDeme object to MeTTa atomspaces, one per deme. Each deme's representation, instance set, and ID are now stored as separate atoms in a dedicated space instead of bundled into a single mkDeme tuple.
Changes:
- CreateDeme creates a new-space per deme and stores Representation, Instance set, and Deme id as separate atoms.
- HillClimbing optimization, Deme expansion, crossover,  New Instances sampling, and Merging Demes now read/write the deme space via space operations instead of passing mkDeme values by return.

## Motivation and Context
Using the mkDeme object forces every function in the deme pipeline to destructure and rebuild the  tuple even when only the instance set changes. Storing each deme in its own space decouples the fields and sets up future parallel deme processing since each deme's state is isolated.

## How Has This Been Tested?
It has been tested  inside a petta environment and testcases have been added. More over, the changes have been tested using the demo problems found in the codebase. 

## Types of changes

## Checklist:

### DIFF
--- a/deme/tests/expand-demes-test.metta
+++ b/deme/tests/expand-demes-test.metta
@@ -1,3 +1,4 @@
+!(import! &self ../demeSpaceName.py)
 !(import! &self ../../utilities/general-helpers)
 !(import! &self ../../utilities/list-methods)
 !(import! &self ../../utilities/pair)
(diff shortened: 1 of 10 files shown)


deme/tests/expand-demes-test.metta
```diff
@@ -443,3 +444,19 @@
 ;; ;;    )
 ;; ;;    $isValid)
 ;; ;;    True)
+
+;; testcase for the expand demes, using deme spaces
```
Yagth, line 448: I wonder why you left these test cases commented out 🤔?

Yagth (approved):
The code looks good to me. But there are some failing test cases. I will merge the PR for now but make sure to address them in the next PR.

deme/tests/expand-demes-test.metta
Yagth: @Liya-F

Liya-F: > I wonder why you left these test cases commented out 🤔?

Oh okay, I will put it inside assertEqual test. Some tests are failing because they are testing the older version methods, I will correct each test or comment out some of them for now. 
Sorry for the late reply I only saw the first comment :)

Yagth: Okay just make sure to add it in the next PR

Issue #471, opened 2026-06-22
Refactor helper functions to recursive impl.


Yagth:
The following functions `getArgs`, `sampleLogicalPerms`, `getMoveArgs`, `sampleStrategyPerms`, `genPairs` and `genTriplet` are using non determinism unnecessarily. According to some testing, non deterministic logic tends to be slower on some occasions than a similar function implemented using recursion or built in higher order deterministic functions. 

The task should refactor this functions into deterministic implementation and document if that shows significant performance change and keep the refactor only if such a gain is observed.

Latest work related to this task can be found on the `ref/RepresentationBuilding` branch for now.

PR #479, opened 2026-06-30
ref(sample-logical-perms): direct-index permutation sampling + randomized FS tie-break

Leul-Negash:
## Description
Two related changes:

1. **Direct-index permutation sampling.** The samplers only need ~arity random pairs/triplets, but we were building the whole `n*(n-1)` / `n*(n-1)*(n-2)` list and then picking a few. Following classic `build_knobs::sample_logical_perms`, each random index now maps straight to its pair/triplet via arithmetic (`pairAt` / `tripletAt`), so the full list is never built. 

2. **Randomized tie-breaking in simple feature selection.** On equal MI scores the selector always kept the lowest-index feature, so every deme picked the same subset on ties. It now uses a seeded comparator (`featureSet<`), matching classic `ScoredFeatureSetGreater`.

## Motivation and Context
 Refs #471 Removes unnecessary non-determinism and the wasteful full-list materialization (much faster on large inputs, identical output), and restores
the feature-set diversity across demes that the deterministic tie-break was
killing.

## How Has This Been Tested?
- Sampling output is byte-for-byte identical to the previous implementation
  (14 seeds, all arities, both edge branches).
- Tie-break: mux6 (simple FS, 3 demes, 5 seeds) was stuck at −24 every run;
  randomized reaches −20 on 4/5 and is never worse.
- All dependent tests pass (sample-perms, knob building, simple FS).

## Types of changes

## Checklist:

### DIFF
--- a/feature-selection/simple.metta
+++ b/feature-selection/simple.metta
@@ -10,14 +10,16 @@
 (= (simpleFeatureSelector $th $numDesired $expDist (mkITable $table $labels) $acc)
     (if (== $numDesired 0)
         ((0 (List.listToExpr (inputLables $labels))))                                                  ;; return all features by default
+
+        (chain (py-call (random.randint 0 2147483647)) $seed                                        ;; random seed to break ties between equally-scored features
         (chain (List.length $labels) $colNum
             (chain (List.length $table) $rowNum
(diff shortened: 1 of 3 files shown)


Leul-Negash:
Splitting this into two focused PRs as discussed direct-index sampling in #480 and the tie-break in #481. Closing this one.

PR #480, opened 2026-06-30
 ref(sample-logical-perms): direct-index permutation sampling

Leul-Negash:
Description

The permutation samplers only need ~arity random pairs/triplets, but we were building the whole n*(n-1) / n*(n-1)*(n-2) list and then picking a few. Following classic build_knobs::sample_logical_perms, each random index now maps straight to its pair/triplet via arithmetic (pairAt / tripletAt), so the full list is never built. genPairs/genTriplet are kept as deterministic recursive utilities.

Motivation and Context

Related to #471 Removes the wasteful full-list materialization — much faster on large inputs, output identical.

How Has This Been Tested?

Output is byte-for-byte identical to the previous implementation (14 seeds, all arities, both edge branches); much faster end-to-end on large inputs. All dependent tests pass (sample-perms, knob building).

Types of changes

Checklist:

### DIFF
--- a/representation/sample-logical-perms.metta
+++ b/representation/sample-logical-perms.metta
@@ -1,133 +1,124 @@
-;; pairInnerLoop -> selects second candidate for each anchor and 
-;;                 builds (anchor partner) pairs.
-;; Type: (-> Expression Number Number Number Expression)
-(= (pairInnerLoop $list $i $j $size)
-    (if (>= $j $size)
-        ()
-        (if (== $i $j)
-            (pairInnerLoop $list $i (+ $j 1) $size)
-            (cons-atom ((index-atom $list $i) (index-atom $list $j))
-                      (pairInnerLoop $list $i (+ $j 1) $size)))))
+;; Permutation sampling for logical / strategy knob building. Only ~arity random
+;; pairs/triplets are ever needed, so rather than enumerating the whole list we
+;; map each random index straight to its pair/triplet (pairAt / tripletAt),
+;; mirroring classic build_knobs::sample_logical_perms.
 
-;; pairOuterLoop -> fixes the first element of the pair (i) and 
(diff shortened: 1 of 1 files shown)


representation/sample-logical-perms.metta
```diff
@@ -1,133 +1,124 @@
...
+    (if (== $picked ())
+        ()
+        (let* (($i (car-atom $picked))
+               ($rest (cdr-atom $picked))
+               (($a $b $c) (tripletAt $n $i))
+               ($move1 (index-atom $moveLabels $a))
+               ($move2 (index-atom $moveLabels $b))
+               ($move3 (index-atom $moveLabels $c)))
+            (cons-atom ($op $move1 $move2 $move3)
+                       (getMoveArgs $op $n $rest $moveLabels)))))
```
Yagth: This function can be optmized and simpler by using `map-atom`

representation/sample-logical-perms.metta
```diff
@@ -1,133 +1,124 @@
...
+        ()
+        (let* (($i (car-atom $picked))
+               ($rest (cdr-atom $picked))
+               (($a $b) (pairAt $n $i))
+               ($1 (index-atom $argLabels $a))
+               ($2 (index-atom $argLabels $b))
+               ($expr (if (< $b $a)
+                          ($swapedOp $2 $1)
+                          ($swapedOp (NOT $1) $2))))
+            (cons-atom $expr (getArgs $swapedOp $n $rest $argLabels)))))
```
Yagth: This implementation can also be refactored and made simpler using `map-atom`. It is easier to read that way

Leul-Negash:
@Yagth I have switched both getArgs and getMoveArgs to use map-atom

Yagth (approved):

PR #489, opened 2026-07-06
Restrict feature selection to the exemplar's mispredicted rows (restrict_incorrect)

Leul-Negash:
## What
Ports classic MOSES `restrict_incorrect` (`feature_selector.cc`). Before feature selection runs, the input table is restricted to the rows the current exemplar mispredicts, so selection favors features that correlate with the exemplar's errors. Falls back to the full table when the exemplar is already correct everywhere, or when the wrong rows share one output value (zero MI).

## Status

Lands the mechanism only. With the `sim` selector it's neutral on `mux6`/`parity3` (classic `simple` also returns a single feature set, so there's no population to differentiate). The payoff needs a population-producing selector (`hc`), which is broken and tracked under a separate issue; once that lands, we can measure whether this improves result quality.

## Testing

`runDemoProblem` with `fsAlgo=sim` on `parity3` runs the new path end to end, reduces cleanly, and returns a valid result. The `fsAlgo=None` path is unchanged.

### DIFF
--- a/representation/representation.metta
+++ b/representation/representation.metta
@@ -86,7 +86,7 @@
 (= (ric-blank-xmplr-fts $fts $rows $len)
    (map-atom $rows $row
       (map-atom (genList 0 (- $len 1)) $i
-         (if (ric-is-xmplr-ft $i $fts) False (index-atom $row $i)))))
+         (if (once (is-member $i $fts)) False (index-atom $row $i)))))
 
 (= (ric-is-xmplr-ft $i $fts)
    (> (size-atom (filter-atom $fts $f (== $f $i))) 0))
--- a/representation/representation.metta
+++ b/representation/representation.metta
@@ -65,27 +65,36 @@
         (representation $nDeme $exemplar (mkITable $table $labels) $ft-selection-algo $prune-exemplar $enforced-fts $scorerType True))
 
 ;; restrictToIncorrect -- classic restrict_incorrect: feed feature selection only
-;; the rows the exemplar mispredicts. Falls back to the full table when it is
-;; already correct everywhere, or when the wrong rows share one output (zero MI).
+;; the rows the exemplar mispredicts, with the exemplar's own columns blanked out.
+;; Falls back to the untouched table when it is already correct everywhere, or when
+;; the wrong rows share one output (zero MI).
 (= (restrictToIncorrect $exemplar (mkITable $table $labels))
    (let* (($expTree (preOrder $exemplar))
           ($len (List.length $labels))
           ($wrongRows (filter-atom $table $row
                         (not (== (scoreTreeHelper $labels $expTree $row)
                                  (index-atom $row (- $len 1)))))))
-     (if (or (== $wrongRows ()) (ric-constant-output $wrongRows $len))
+     (if (ric-constant-output $wrongRows $len)
          (mkITable $table $labels)
-         (mkITable $wrongRows $labels))))
+         (mkITable (ric-blank-xmplr-fts (treeFtsIndices $exemplar $labels) $wrongRows $len) $labels))))
 
-;; True when every row's output column (last column) holds the same value
+;; On the rows the exemplar gets wrong its own features still track the output, so
+;; feature selection would just hand back the features the exemplar already uses and
+;; the search would never gain a new one. Classic blanks those columns rather than
+;; dropping them: a constant column carries no mutual information so it is never
+;; selected, and every remaining column keeps its index.
+(= (ric-blank-xmplr-fts $fts $rows $len)
+   (map-atom $rows $row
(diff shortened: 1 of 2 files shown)


representation/representation.metta
```diff
@@ -64,6 +64,29 @@
...
+
+;; True when every row's output column (last column) holds the same value
+(= (ric-constant-output $rows $len)
+   (ric-all-same (map-atom $rows $r (index-atom $r (- $len 1)))))
+(= (ric-all-same $xs)
+   (if (== $xs ()) True (ric-all-eq (car-atom $xs) (cdr-atom $xs))))
+(= (ric-all-eq $h $xs)
+   (if (== $xs ())
+       True
+       (if (== (car-atom $xs) $h) (ric-all-eq $h (cdr-atom $xs)) False)))
```
Yagth:
```suggestion
(= (ric-constant-output $rows $len)
    (<= (size-atom (unique-atom (map-atom $rows $r (index-atom $r (- $len 1))))) 1))
```
Can we not simplify this like so?

Yagth (changes requested):
Looking good and most thing appear similar to classic.
But when I was comparing scores against the base branch on the parity3 and mux6 and unfortunately both regress significantly.
 - parity3 (--fsAlgo=sim --seed=0 --maxGen=50 --nDeme=1): base solves it (score 0, done by generation 14); this branch never gets above -4 in all 50 generations.
 - mux6 (--maxGen=10): base finishes 10 generations in ~20s; this branch freezes in generation 2 — no result even after 19 minutes, reproduced on four runs. Can you check from your side one last time?

I would also appreciate it if you include a simple smoke test case for your change. Better yet, if you can, a test case from the classic implementation.

Leul-Negash:
> Looking good and most thing appear similar to classic. But when I was comparing scores against the base branch on the parity3 and mux6 and unfortunately both regress significantly.
> 
> * parity3 (--fsAlgo=sim --seed=0 --maxGen=50 --nDeme=1): base solves it (score 0, done by generation 14); this branch never gets above -4 in all 50 generations.
> * mux6 (--maxGen=10): base finishes 10 generations in ~20s; this branch freezes in generation 2 — no result even after 19 minutes, reproduced on four runs. Can you check from your side one last time?
> 
> I would also appreciate it if you include a simple smoke test case for your change. Better yet, if you can, a test case from the classic implementation.

@Yagth you were right on both, I reproduced them exactly. parity3 sat at -4 and mux6 froze in generation 2.

The row filter was fine. The problem is I ported restrict_incorrect on its own. On the rows the exemplar gets wrong its own features still track the output, so sim handed the same features straight back, nothing new entered the representation and the search stalled. Classic never runs it alone: feature_selector.h defaults restrict_incorrect(true) alongside ignore_xmplr_features(true), and feature_selector.cc explains why, "one can ignore the features in the exemplar during feature selection. That way, fs will never pick features in the exemplar." It does that by blanking those columns rather than dropping them, so a blanked column has zero MI and the rest keep their index. I do the same now, and the fallback paths return the untouched table so behaviour matches base whenever the restriction does not apply.

parity3 with your exact flags (--fsAlgo=sim --seed=0 --maxGen=50 --nDeme=1): score 0, stops after 14 generations, same as base.

mux6 I ran as --fsAlgo=sim --maxGen=10, since with the default None the restricted path is never reached: finishes at -16 in ~164s, base gets -20 in ~129s.

Applied your unique-atom suggestion. It also let the (== $wrongRows ()) check go, since an empty row set already passes <= 1.

I could not port a test from classic because restrict_incorrect has no coverage there. It only appears in feature_selector.h, feature_selector.cc and problem-params.cc, nothing under tests/. So I wrote three cases in representation-test.metta from the three branches in feature_selector.cc: restrict when the wrong rows carry different outputs, keep the full table when the exemplar is correct everywhere, keep it when the wrong rows share one output.

representation/representation.metta
```diff
@@ -64,6 +64,38 @@
...
+
+;; On the rows the exemplar gets wrong its own features still track the output, so
+;; feature selection would just hand back the features the exemplar already uses and
+;; the search would never gain a new one. Classic blanks those columns rather than
+;; dropping them: a constant column carries no mutual information so it is never
+;; selected, and every remaining column keeps its index.
+(= (ric-blank-xmplr-fts $fts $rows $len)
+   (map-atom $rows $row
+      (map-atom (genList 0 (- $len 1)) $i
+         (if (ric-is-xmplr-ft $i $fts) False (index-atom $row $i)))))
```
Yagth:
```suggestion
         (if (once (is-member $i $fts)) False (index-atom $row $i)))))
```

Leul-Negash:
@Yagth change applied

Yagth (approved):

PR #492, opened 2026-07-11
Optimize crossover helpers

Yadelew-zemene:
## Description
Optimizes the crossover implementation without changing behavior or output.

### Main changes
* **Performance Refactor:** Refactored the pairwise and triple crossover helpers to thrack and thread the current offspring count through recursion instead of repeatedly computing `List.length on the generated instances`. This removes unnecessary linear scans while strictly preserving traversal order.
* **Module Renaming:** Renamed the crossover modules and updated imports to reflect that they now contain the complete crossover implementation rather than only `crossTopOne`.
* **Test & Helper Updates:** Updated helper implementations and unit tests to match the new helper return values while preserving existing crossover behavior.

## Motivation and Context
This change is a performance and organazinig refactor. It removes unnecessary work performed during offspring generation while keeping the crossover algorithms, generated offspring, and C++ traversal behavior unchanged. The file renaming also improves organization by making the module structure match the current implementation.

## How Has This Been Tested?
* Updated helper unit tests for the new helper return values.
* Verified existing crossover tests continue to pass.
* Confirmed no behavioral changes to the generated offspring or crossover traversal.

## Types of changes
- [x] Performance improvement (non-functional refactor)

## Checklist
- [x]  I have added tests to cover my changes


optimization/hillclimbing/test/crossover-test.metta
```diff
@@ -118,30 +118,47 @@
...
+            (cons
+              (mkSInst (mkPair (mkInst (cons 2 ())) (mkCscore 0 0 0 0 3)))
+              ()))
+          (getInst (mkSInst (mkPair (mkInst (cons 5 ())) (mkCscore 0 0 0 0 3))))
+          (mkInst (cons 9 ()))
+          ()
+          0
+          2))
+    )
+    ((List.length $instances) $count)
```
Yagth: Hey yadelew, I prefer the exact match test case instead of the count here. It helps to catch when something changes even though it is a bit painful to maintain.

Yagth (commented):
Hey yadelew. The changes looks good to me. The only change I would like to see is  a more concrete test cases (exact returned result matches) instead of the current count based test case matches.

Yadelew-zemene:
> Hey yadelew. The changes looks good to me. The only change I would like to see is a more concrete test cases (exact returned result matches) instead of the current count based test case matches.

Ok. i will change.

optimization/hillclimbing/test/crossover-test.metta
Yadelew-zemene: > Hey yadelew, I prefer the exact match test case instead of the count here. It helps to catch when something changes even though it is a bit painful to maintain.

Test updated accordingly. As the change is only for performance reason and no logic change, it does not leads to the complex refacting  through the crossover pipeline. Right.

Yagth: Are you sure you pushed the changes? Because I am still seeing the old version I asked to be changed?

Yadelew-zemene: > Are you sure you pushed the changes? Because I am still seeing the old version I asked to be changed?

Please could  you check it  now?

Yagth (approved):

Issue #487, opened 2026-07-06
Hill climber based feature selection refactor


Yagth:
The hill climber based feature selection algorithm has fallen behind a lot. This algorithms needs to be up to date with our latest code changes and it's test cases pass.

PR #493, opened 2026-07-14
ref(hc): fix the feature-selection import order and make the precision scorer reachable

Leul-Negash:
## Description

Makes the hill-climbing feature selector work, and makes the precision scorer selectable.

**Import order.** `hc.metta` carried two late-binding workarounds whose only purpose was to defer calls into modules imported after it; both are removed. The edge closing the cycle was `bitsToIndices`, which lived in `deme/score-deme.metta` while being needed during feature selection, so it moves to `feature-selection/feature-selection-helpers.metta`. `hc` and the representation modules are now imported after `optimization/hillclimbing`.

**A missing import.** `scoring/precision-bscore` was not imported by `moses.metta` at all. The hardcoded `mi` hid this: once the scorer type actually pointed at precision, the run failed in Prolog with a type error.

**Scorer selection.** The scorer is chosen through a new `fsScorerType` parameter (`mi` or `pre`). The five scorer hyperparameters were hardcoded in two places; they are now defaults in `parameters/defaults.metta` and pick up CLI flags through the existing generic flag parsing. `hcFtSetPop` takes the scorer, and the five as a single fourth parameter that it appends the scorer to in its own body, so `featureSelector` and `featureSelection` no longer thread the five through their signatures.

**Initial features.** The hc path ignored `$init-fts` and always started from an all-zero instance. Classic turns on one bit per initial feature (`initial_instance` in `feature-selection/algo/deme_optimize.cc`). `indicesToBits`, the inverse of the existing `bitsToIndices`, is added and builds the instance. With no initial features it yields the same all-zero instance as before, so current runs are unaffected.

**Multiple demes.** `createRepresentation` and `runMoses` each destructured their result with a one-element pattern, so any feature selector returning more than one set (`nDeme > 1`) collapsed to Empty before a deme could expand. `createRepresentation` now returns the representation list directly, and `runMoses` reports the deme count instead of pulling apart a single deme. `nDeme = 1` is unchanged; `nDeme > 1` now runs end to end.

**`normalizeFsHyperparams`.** It fills in the defaults for callers that hand `mkParams` a bare scorer name or `()`, which is how the hillclimbing and simulated annealing tests reach the feature selection path. It recognized a bare name by testing `get-metatype` for `Symbol`, but `mi` is also the name of a similarity scorer, so it comes back `Grounded`, the test never fired, and `mi` was handed back unchanged and reduced to Empty while `pre` worked. It tests for `Expression` instead.

**Types.** `getHcFtPop`'s signature is left commented out rather than restated. The declared `(List ...)` types never match the flat `cons` tuples the function returns, so any call reduces to Empty.

**Tests.** `hc-testold.metta` is deleted. It could not have run: the filename does not match the `*test.metta` glob used by `scripts/run-tests.py`, and it binds its table with `bind!`, whose symbols do not match an `mkITable` pattern. `hc-test.metta` replaces it.

`bestPossibleScore` remaining 1.0 for feature selection is left as is.

Review order:

1. `moses.metta` — the import list, which is the core of the change.
2. `feature-selection/hc.metta`, `deme/score-deme.metta`, `feature-selection/feature-selection-helpers.metta` — the `bitsToIndices` move that made the ordering possible, and its new inverse.
3. `optimization/hillclimbing/hill-climbing-helpers.metta` — `defaultFsHyperparams` and `normalizeFsHyperparams`.
4. `parameters/defaults.metta`, `feature-selection/feature-selection-main.metta`, `representation/representation.metta`, `deme/expand-deme.metta` — the parameter plumbing.
5. `representation/create-representation.metta`, `deme/expand-deme.metta` — the one-element destructures dropped so `nDeme > 1` runs.
6. `feature-selection/tests/` — the new coverage.

## Motivation and Context

Addresses #487. The hill-climbing feature selector could be chosen with `--fsAlgo=hc` but did not work, and the precision scorer could not be reached at all, which in turn hid the fact that its module was never imported.

## How Has This Been Tested?

- Full suite via `scripts/run-tests.py`: 84 files, all passing.
- `feature-selection/tests/hc-test.metta`: 5 assertions. The table is `Output = A OR C`, so the winning feature set has to be `(0 2)`. They cover `mi`, `pre`, that `miConfi` reaches the scorer, and that initial features reach the search. `mi` and `pre` are asserted to give different runner-up scores deliberately: that is what catches a silent fall back to mutual information through the `case` else branch in `featureSetScorer`.
- `feature-selection/tests/feature-selection-helpers-test.metta`: `indicesToBits`, including that it round-trips through `bitsToIndices`.
- End to end: `moses.metta --problem=mux6 --fsAlgo=hc` runs to completion under both `--fsScorerType=mi` and `--fsScorerType=pre`, reaching the same score on both.
- Multiple demes: `--fsAlgo=hc --nDeme=3` runs to completion and returns a result, where it previously reduced to Empty; `--nDeme=1` is unchanged.

## Types of changes

## Checklist:


optimization/hillclimbing/hill-climbing-helpers.metta
```diff
@@ -192,6 +192,27 @@
     ($rows $labels)
 )
 
+;; The configured feature selection hyper parameters, in the order the scorers read them.
+(= (defaultFsHyperparams)
+   ((param fsMiConfi) (param fsPrePenalty) (param fsPreMinActivation) (param fsPreMaxActivation) (param fsPrePositive)))
+
+;; Appends the scorer name, which is what the scorers expect in the last slot.
+(= (withFsScorer $fsHyperparams $scorerType)
+   (union-atom $fsHyperparams (cons-atom $scorerType ())))
```
Yagth: Why are we wrapping the `$scorerType` in a `cons-atom`? Isn't this equivalent to just `($scorerType)`?

optimization/hillclimbing/hill-climbing-helpers.metta
```diff
@@ -192,6 +192,27 @@
...
+;;   ()                              -- defaults, scored by mutual information
+;;   a scorer symbol, mi or pre      -- defaults, scored by that scorer
+;;   the full hyper parameter tuple  -- used as is
+;; A scorer name is matched on metatype rather than on Symbol: `mi` is also the
+;; name of a similarity scorer, which makes it Grounded instead of Symbol.
+(= (normalizeFsHyperparams $fsScorer)
+   (if (== (get-metatype $fsScorer) Expression)
+       (if (== $fsScorer ())
+           (withFsScorer (defaultFsHyperparams) mi)
+           $fsScorer)
```
Yagth, line 215: I am not sure I understand why the `$fsScorer` is the only returned value here? Doesn't it need the default hyperparameters as along with it?

feature-selection/hc.metta
```diff
@@ -1,21 +1,29 @@
...
-(= (getHcFtPop (cons (mkSInst (mkPair (mkInst $list) $score)) $sInsts) $acc) 
-  (chain (getHcFtPop $sInsts $acc) $new-acc
-    (chain ($score (bitsToIndices $list 0 ())) $score-ft-set-pair
-      (cons $score-ft-set-pair $new-acc))))
+(= (getHcFtPop (cons (mkSInst (mkPair (mkInst $list) $score)) $sInsts) $acc)
+  (let* (($new-acc (getHcFtPop $sInsts $acc))
+         ;; Reduced on its own line: inside ($score (bitsToIndices ...)) the pair's
+         ;; head is a number, so the call would never be evaluated.
+         ($ft-set (bitsToIndices $list 0 ()))
+         ($score-ft-set-pair (cons-atom $score (cons-atom $ft-set ()))))
```
Yagth: Same question here. Isn't this the same as? `(cons-atom $score ($ft-set))` which is in turn the same as `($score $ft-set)`, why the workaround of using cons-atom?

Yagth (commented):
Looking good. Just few questions before merge.

optimization/hillclimbing/hill-climbing-helpers.metta
Leul-Negash: not equivalent unfortunately. mi is grounded since it's also the name of a similarity scorer in similarity-scorers.metta, so ($scorerType) gets evaluated as a call. i get (cons 50.0 (cons 1.0 (cons 0.5 (cons 1.0 (cons true (partial mi ())))))) back instead of a flat tuple. pre is a plain symbol so it works there, which is what makes it easy to miss, it'd be fine for pre and silently wrong for mi. the cons-atom just keeps the name as data. same collision as the metatype check right below.

optimization/hillclimbing/hill-climbing-helpers.metta
Leul-Negash: that branch is the case where the caller already passed the full tuple, so it's complete and adding the defaults would overwrite what they set. bare scorer names fall to the else branch and do get them. i checked all four shapes, () gives (50.0 1.0 0.5 1.0 true mi), mi gives the same, pre gives (50.0 1.0 0.5 1.0 true pre), and (0.001 1.0 0.5 1.0 True pre) comes back unchanged.

feature-selection/hc.metta
Leul-Negash: @Yagth yeah you're right on this, that one's just a workaround with no reason behind it. $score is a number so the head never gets applied and it stays data. simplified it to ($score $ft-set), same output including the (0.0 ()) case, and hc-test still passes. kept $ft-set on its own line though, that one does need it since bitsToIndices wouldn't evaluate inside the pair.

optimization/hillclimbing/hill-climbing-helpers.metta
Yagth: Are you familiar with the function `quote`? You can use that to prevent such automatic evaluations.

Leul-Negash: yeah that's cleaner, switched to it.

Yagth (approved):

PR #495, opened 2026-07-21
feat: store input tables in column-major format

Yadelew-zemene:
## Description
This PR changes the internal representation of input tables from **row-major** to **column-major** to match how they are used throughout the MOSES evaluation pipeline. The implementation updates the CSV loading pipeline, input table construction, and the dependent utilities and tests to work with the new representation while preserving existing behavior.

The main changes include:
- Store input tables in column-major format instead of row-major.
- Update the CSV loader and input table construction accordingly.
- Update functions that assumed row-major layout to operate on the new representation.
- Update related unit tests and benchmark files.
- Preserve existing functionality while reducing the overhead of repeatedly converting between row-major and column-major layouts.

## Motivation and Context
Most consumers of the input table access data by **column**, while the existing representation stores it by **row**. This causes repeated row-to-column transformations throughout execution, introducing unnecessary overhead.

By storing the table directly in column-major format, these conversions are eliminated, reducing runtime while keeping the external behavior unchanged.

## How Has This Been Tested?
- Updated the affected unit tests to use the new input table representation.
- Verified that all existing tests pass.
- Benchmarked using the **mux6** benchmark with:
  - `nEval = 1000`
  - `maxGen = 1`
  - 5 runs before and 5 runs after the change.

Average wall-clock (`real`) execution time:

| Before | After | Improvement |
|--------:|------:|------------:|
| 57.4 s | 43.8 s | **~24% faster** |

## Types of changes

## Checklist:

### DIFF
--- a/feature-selection/feature-selection-helpers.metta
+++ b/feature-selection/feature-selection-helpers.metta
@@ -20,37 +20,36 @@
             (List.uniqueValuesCount $xs $new-tuple-added))))
 
 ; (: Table.getColumn (-> Number (List (List $a)) (List $a)))
-(= (Table.getColumn $i ()) ())
-(= (Table.getColumn $i (cons $row $rest))
-    (cons (List.at $i $row) (Table.getColumn $i $rest)))
-; ;; Table.pop        -- gets and removes a column from the current ttable structure
-; ;;                  -- returns the remaining table and the culumn asked for
-(= (Table.pop $i ()) ())
-(= (Table.pop $i (cons $row $rest))
-    (let (cons $d $ds) $row
-        (if (== $i 0)
-            (cons $ds (Table.pop $i $rest))
-            (cons (cons $d (List.pop (- $i 1) $ds)) (Table.pop $i $rest)))))
-; removes the first column from table
-; ; (: Table.pop (-> (List (List $a)) (List (List $a))))
-; (= (Table.pop $rows)
-;     (unify $rows () ()
-;         (unify $rows (cons (cons $d $ds) $rest)
-;             (cons $ds (Table.pop $rest)) ())))
+(= (Table.getColumn $i $table)
+    (if (== $table ())
+         ()
+       (List.at $i $table)))
 
-;; returns the last column of the table
+;; Table.pop        -- gets and removes a column from the current ttable structure
+;;                  -- returns the remaining table
+(= (Table.pop $i $table)
+    (if (== $table ())
+         ()
+         (List.pop $i $table)))
+
+;; Removes the first column from a column-major table.
+; (: Table.pop (-> (List (List $a)) (List (List $a))))
+(= (Table.pop $table)
+    (List.pop 0 $table))
+
+;;  returns the last column of the table
 ; (: Table.getOutputColumn (-> ITable (List $a)))
 (= (Table.getOutputColumn (mkITable $table $labels))
-    (chain (List.length $labels) $colNum 
-    (chain (Table.getColumn (- $colNum 1) $table) $oc   
-    $oc  ))
-)
+    (chain (List.length $labels) $colNum
+        (Table.getColumn (- $colNum 1) $table)))
+
 
 ;;;;;;;;;;;;;;
 ;; Transpose;;
 ;;;;;;;;;;;;;;
 ; (: Table._transposeHelper (-> (List Number) (List (List $a)) (List (List $a))))
 (= (Table._transposeHelper () $table) ())
+
 (= (Table._transposeHelper (cons $i $rest) $table)
     (cons (Table.getColumn $i $table)
           (Table._transposeHelper $rest $table)))
@@ -236,7 +235,7 @@
             (if (= $datarows (cons $row $rows))
                 (chain (map-atom $indices $i (List.at $i $row)) $dataTuple
                     (cons $dataTuple (zipColumn $indices $rows)))
-                ()))))
+                 ()))))
 
 ;; return marginal probabilities as tuple (p1 p2)
 ; (: marginalProbs (-> Number Expression Expression))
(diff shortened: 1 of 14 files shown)


moses/tests/input-file-test.metta
```diff
@@ -105,4 +104,4 @@
 !(assertEqual
    (let $result (moses)
         (second (index-atom (second $result) 0)))
-   0)
+-1)
```
Yagth, line 107: I wonder why the score changed here?

feature-selection/feature-selection-helpers.metta
```diff
@@ -222,21 +221,29 @@
                     ;; feature MI calculation
             (chain (zipColumn $featureIndices $rows) $jointColumn
                 (chain (mutualInformation $jointColumn $targetColumn) $miScore
-                                                                                            ;; the default confidence scaling parameter
-                        (chain (calculateConfidence (size-atom $featureIndices) (List.length $rows) 50.0) $confidence
-                                (* $miScore $confidence)))))))
+                    (chain (List.length $targetColumn) $rowCount ;; Use the number of examples for confidence scaling.
```
Yagth: This was fine the way it was. Why define the variable using chain?

feature-selection/feature-selection-helpers.metta
```diff
@@ -222,21 +221,29 @@
...
+; (: zipColumn (-> Expression (List (List $a)) (List $a)))
+(= (zipColumn $indices $table)
     (if (== $indices ())
         ()
-        (unify $datarows () ()
-            (if (= $datarows (cons $row $rows))
-                (chain (map-atom $indices $i (List.at $i $row)) $dataTuple
-                    (cons $dataTuple (zipColumn $indices $rows)))
-                ()))))
+        (chain (map-atom $indices $i (Table.getColumn $i $table)) $columns
```
Yagth: Using `let*` is actually better for these. It is easily readable.

Yagth (changes requested):
The test cases of the feature-selection algorithms has changed, I don't think a change in table representation should result in such difference. Can you correct it?

Yadelew-zemene:
> The test cases of the feature-selection algorithms has changed, I don't think a change in table representation should result in such difference. Can you correct it?

Good. I think first in that way also. But after keeping track of function call, there are still some helpers still assume row based table in feature selection pepipline.So  i suspect for that converting to column wise table may results in different feature selection and as you saw  the differece is  just swap between (0 and -2, -1.5 and -3.5),not generate another number and we can take this  as sign that it does not select the features as intended.

moses/tests/input-file-test.metta
Yadelew-zemene: > I wonder why the score changed here?

All right. There are some helpers in bscore that still assumes row major. These may affect the scoring context.

feature-selection/feature-selection-helpers.metta
Yadelew-zemene: > This was fine the way it was. Why define the variable using chain?

Ok. But as we are now using the length of column to get the number of rows , it is preferable to in introduce the variable that  represent this context rather that left it as it was.Right.
If you prefer the original i can switch it to that.

feature-selection/feature-selection-helpers.metta
Yadelew-zemene: > Using `let*` is actually better for these. It is easily readable.

Ok . i will switch to that.

Yagth:
> > The test cases of the feature-selection algorithms has changed, I don't think a change in table representation should result in such difference. Can you correct it?
> 
> Good. I think first in that way also. But after keeping track of function call, there are still some helpers still assume row based table in feature selection pepipline.So i suspect for that converting to column wise table may results in different feature selection and as you saw the differece is just swap between (0 and -2, -1.5 and -3.5),not generate another number and we can take this as sign that it does not select the features as intended.

That's not the change I was talking about actually. I am more worried about the change in mutual information while the features are the same. For example in the `similarity-scorers-test.metta` file line number 22 the mutual information for the same features `(0 1)` and `(0 3)` was 0.815 now it has increased to 0.9. This change is undesirable.

feature-selection/feature-selection-helpers.metta
Yagth: That doesn't make a difference. I mean it is a small edit but instead of adding more chain calls, we can simplify the code by in-lining the `(list.length ...)` function call instead of the `$rowCount` variable (as this variable isn't used anywhere).

Yadelew-zemene: > That doesn't make a difference. I mean it is a small edit but instead of adding more chain calls, we can simplify the code by in-lining the `(list.length ...)` function call instead of the `$rowCount` variable (as this variable isn't used anywhere).

The requested changes  addressed accordingly

Yadelew-zemene:
> > > The test cases of the feature-selection algorithms has changed, I don't think a change in table representation should result in such difference. Can you correct it?
> > 
> > 
> > Good. I think first in that way also. But after keeping track of function call, there are still some helpers still assume row based table in feature selection pepipline.So i suspect for that converting to column wise table may results in different feature selection and as you saw the differece is just swap between (0 and -2, -1.5 and -3.5),not generate another number and we can take this as sign that it does not select the features as intended.
> 
> That's not the change I was talking about actually. I am more worried about the change in mutual information while the features are the same. For example in the `similarity-scorers-test.metta` file line number 22 the mutual information for the same features `(0 1)` and `(0 3)` was 0.815 now it has increased to 0.9. This change is undesirable.

The test failure for the previous  tests was due to its assumption of row major for calculating the table length.
After changing this to length of column, the previous tests  have restored correctely.

Yagth (approved):

PR #497, opened 2026-07-22
Feat/santa fe problem

K1rubel:
## Description

This PR builds on the completed action-reduction framework to implement the full Santa Fe ant problem in MeTTa MOSES—from action-tree execution and scoring through representation, hill climbing, metapopulation search, and deterministic perfect-score convergence.

## Changes

- Added the 32×32 Santa Fe trail, toroidal movement, food perception, stateful action execution, and 89-pellet scoring.

- Added action-tree complexity and candidate reconstruction using cleanActionTree.

- Wired problem=ant into the normal moses-run dispatcher.

- Added randomized action representations and multi-deme construction.

- Added an ant-specific search lifecycle with:
     - actual scorer-call accounting;

     - incremental and duplicate-aware scoring;

    - visit-limited exemplar selection;

    - configurable dominance filtering;

    - distinct expansion IDs;

    - separate raw-score champion and penalized-score archive;

    - global and per-deme evaluation limits.

- Aligned neighborhood sampling and crossover behavior with classic C++ MOSES.

- Added domain, integration, hill-climbing, sampling, and end-to-end pipeline tests.

- Documented differences from Boolean evolution and a future policy-based unification path.

## Result

The deterministic seed-1 test reaches raw score 0, confirming that the evolved tree eats all 89 pellets. It converges through expansion 2 after 520 actual ant simulations.

## What fundamentally has changed from Boolean counterpart 
(needs further verification)
- Counts actual new scorer calls instead of nominal samples or population size.
- Retains existing scores and avoids rescoring duplicate candidates.
- Tracks the best raw-score candidate independently from the penalized-score archive.
- Uses raw score for target termination while retaining penalized score for optimization and selection.
- Prevents unlimited re-expansion of the same exemplar through visit accounting.
- Preserves dominated candidates as potential evolutionary stepping stones.
- Advances expansion IDs correctly and handles empty representations without consuming a generation.
- Enforces a complete-run evaluation budget across demes.
- Corrects neighborhood sizing, sampling boundaries, and crossover bookkeeping.
- Ensures the raw champion survives archive trimming and appears in final results.
## Motivation and Context
To verify that the action selection branch does useful stuff.

## How Has This Been Tested?

- [x] Santa Fe domain tests
- [x] Ant integration tests
- [x] Perfect-convergence pipeline test

## Types of changes

## Checklist:


representation/tests/representation-test.metta
```diff
@@ -49,6 +63,64 @@
...
+              0))
+  (mkTree (mkNode NOT) ((mkTree (mkNode A) ()))))
+
+!(assertEqual
+  (getCandidateHelper
+      (mkInst (1))
+      (mkKnob (mkNullVex ((mkTree (mkNode D) ())))
+              (mkLSK (mkDiscKnob (mkMultip 3) (mkDiscSpec 0) (mkDiscSpec 0) ((mkDiscSpec 1))))
+              0))
+  (mkTree (mkNode NOT) ((mkTree (mkNode D) ()))))
```
Yagth, line 99: IS this the correct return type? I thought mkInst 1 returns the tree as is. In this case `(mkTree (mkNode D) ())`

optimization/hillclimbing/hill-climbing-helpers.metta
```diff
@@ -1,12 +1,24 @@
-;; Verbatim port of the Infromation theoretic bits function from c++
+;; A negative value means that a direct hill-climbing caller has not supplied
+;; an outer ant evaluation allowance. expandAntDemeHelperCounted temporarily
+;; replaces it with the budget remaining in the complete ant run.
+!(bind! &antLocalEvalLimit (new-state -1))
```
Yagth: You can use the parameters library for this kind of global parameter definitions now. You can define a default in the `parameters/defaults.metta` then get it using the `param` function; set it using `setParam` function.

optimization/hillclimbing/hill-climbing-helpers.metta
```diff
@@ -163,6 +198,88 @@
...
+   (if (== $tree $cachedTree)
+       $cachedScore
+       (antCachedScoreForTree $tree $rest)))
+
+;; Score an already materialized and reduced action tree without reconstructing
+;; it a second time through antBasedScorer.
+(= (scoreCleanedAntTree
+      $cleanedTree
+      (mkAntContext $steps $complexityRatio $actions $perceptions))
+   (getAntCscore $cleanedTree $complexityRatio $steps))
```
Yagth: I couldn't find the definition of this function. Can you point me to it?

optimization/hillclimbing/crossover.metta
```diff
@@ -13,15 +13,18 @@
 
        ($nToMakeNew (if (< (- $sampleSize 1) $nToMake) (- $sampleSize 1) $nToMake))
        (($prevInst $newInst) (List.splitAt $sampleStart $instSet))
-       ($list (List.partialSort instance> $newInst $nToMakeNew ()))
+       ;; Restrict partial_sort to the recorded sample window and request one
+       ;; reference plus nToMake targets.
+       ($sample (List.takeN $sampleSize $newInst))
+       ($list (List.partialSort instance> $sample (+ $nToMakeNew 1) ()))
```
Yagth, line 19: I don't thinks this is what we want to do? Isn't this restricting the sorting to be applied to only the first `$sampleSize` part of the list? We instead want to sort the entire list `$sampleSize` times. The way I see it those two does different things.

optimization/hillclimbing/crossover-helpers.metta
```diff
@@ -96,60 +96,76 @@
...
-           ($newProcessed (List.append $i $processed))
-         )
-         (crossTriples $newProcessed $iRest $base $updatedAcc $limit))))
+;; makeCrossoverSlots -- Model the C++ resize(num_to_make) destination array.
+;; Every reachable slot is populated before it is returned; the sentinel atoms
+;; make an accidental hole visible in a regression instead of hiding it.
+(= (makeCrossoverSlots $count)
+   (if (<= $count 0)
+       ()
+       (cons UnfilledCrossoverSlot (makeCrossoverSlots (- $count 1)))))
```
Yagth, line 138: Why do we need to put a place holder before making a crossover? So far we have only been doing a merging once the cross over is done, not using any form of place holders?

optimization/hillclimbing/hill-climbing-helpers.metta
```diff
@@ -17,15 +29,23 @@
...
 ; (: estimateNeighborhood (-> Number (MultiMap (DiscSpec $knob)) Number))
 (= (estimateNeighborhood $distance $knobMap)
    (if (== $distance 0)
        1
-       (let $nnEstimate (informationTheoreticBits $knobMap)
-          (if (== $distance 1)
+       ;; C++ stores the floating-point information sum in size_t before using
+       ;; it as a neighborhood dimension. `floor-math` is the exact
+       ;; nonnegative truncation and remains grounded inside this `let`.
+       (let $nnEstimate (floor-math (informationTheoreticBits $knobMap))
```
Yagth, line 41: Moving the floor into the informationTheorticBits is more preferable as it also makes the test cases far easier to read.

optimization/hillclimbing/test/hillclimbing-helpers-test.metta
```diff
@@ -72,34 +72,45 @@
...
-!(assertEqual (informationTheoreticBits (knobmap2)) 5)
-;; Test informationTheoreticBits with a knob 
-;; map containing two knobs with same multiplicity of 3
-!(assertEqual (informationTheoreticBits (knobmap1)) 6)
-; ;; Test informationTheoreticBits with a knob 
-; ;; map containing two knobs with same multiplicity of 3
-!(assertEqual (informationTheoreticBits (knobmap1)) 6)
+!(assertEqual
+  (informationTheoreticBits (knobmap1))
+  (+ (log-math 2 3) (log-math 2 3)))
```
Yagth, line 80: I prefer an exact value here in the test case. It is hard to compute log on the fly

optimization/hillclimbing/test/hillclimbing-helpers-test.metta
```diff
@@ -72,34 +72,45 @@
...
-!(assertEqual (estimateNeighborhood 2 (knobmap1)) 15)
-!(assertEqual (estimateNeighborhood 2 (knobmap2)) 10)
-!(assertEqual (estimateNeighborhood 3 (knobmap1)) 20)
-
-!(assertEqual (estimateNeighborhood 7 (knobmap1)) 0)
+;; C++ truncates the information sum to size_t: floor(log2(2)+log2(3))=2
+;; and floor(2*log2(3))=3 before estimating these neighborhoods.
+!(assertEqual (estimateNeighborhood 1 (knobmap2)) 4)
+!(assertEqual (estimateNeighborhood 2 (knobmap1)) 3)
+!(assertEqual (estimateNeighborhood 2 (knobmap2)) 1)
```
Yagth, line 98: This test case is a bit odd. Why does the estimate drop when the distance increase? Shouldn't it at least be as equal as the estimate generated at distance 1?

optimization/hillclimbing/test/hillclimbing-helpers-test.metta
```diff
@@ -72,34 +72,45 @@
...
-!(assertEqual (estimateNeighborhood 1 (knobmap2)) 10)
-!(assertEqual (estimateNeighborhood 2 (knobmap1)) 15)
-!(assertEqual (estimateNeighborhood 2 (knobmap2)) 10)
-!(assertEqual (estimateNeighborhood 3 (knobmap1)) 20)
-
-!(assertEqual (estimateNeighborhood 7 (knobmap1)) 0)
+;; C++ truncates the information sum to size_t: floor(log2(2)+log2(3))=2
+;; and floor(2*log2(3))=3 before estimating these neighborhoods.
+!(assertEqual (estimateNeighborhood 1 (knobmap2)) 4)
+!(assertEqual (estimateNeighborhood 2 (knobmap1)) 3)
```
Yagth, line 97: These test cases have overall dropped in the estimated neighborhood. I understand that is because of the change to the log function which will lower the estimate as compared to the previous simple addition. But were you able to justify the unusually low neighborhood estimates? I expected them to be higher.

Yagth (approved):

representation/tests/representation-test.metta
K1rubel: The `((mkDiscSpec 1))` is list of disallowed settings. So `as is` is disallowed.

optimization/hillclimbing/hill-climbing-helpers.metta
K1rubel: Such problem specific parameter settings have been moved and merged as global setting to the common file.

optimization/hillclimbing/crossover.metta
K1rubel: this is the C++ line that does the same thing as above
```
std::partial_sort(
    deme.begin() + sample_start,
    deme.begin() + sample_start + num_to_make + 1,
    deme.begin() + sample_start + sample_size,
    std::greater<deme_inst_t>());
```

- Consider exactly sample_size candidates.
- Sort only the best num_to_make + 1 candidates into the front.
- The first is the reference candidate.
- The remaining num_to_make candidates are crossover targets.
- The rest of the sample does not need to be ordered.
But I have optimized the partial sorting functionality. I've put comments along the changed lines.

optimization/hillclimbing/hill-climbing-helpers.metta
K1rubel: I think you loose some precision points and the cumulative effect will be big if you truncate the results before they are added as it happens in `fold-atom` that gets to call on `(knobInformationBits $knob)` here
```
(= (informationTheoreticBits $knobMap)
   (foldl-atom (MultiMap.values $knobMap) 0 $bits $knob
     (+ $bits (knobInformationBits $knob))))
```

optimization/hillclimbing/test/hillclimbing-helpers-test.metta
K1rubel: This is exact-distance shell measure; meaning counts configurations that differ from the center in exactly d positions—not up to d positions. So it is not a cumulative sum thing and this is not a monotonic. The binomial estimates shrink beyond the mid point. As an illustration
```
For three binary knobs around (0,0,0):
Distance 0: (0,0,0) → 1 configuration
Distance 1: change exactly one knob → 3
Distance 2: change exactly two knobs → 3
Distance 3: change all three knobs → 1
```

optimization/hillclimbing/crossover-helpers.metta
K1rubel: This is an addition made to match C++ three-way crossover exactly. Same index positions could be generated more than once, meaning overwritting exists. But the MeTTa version just appends childern, keeping those that might have been overwritten by C++ version, because MeTTa lists have no writable empty positions, this would not work. It can be improved but that is the point.

optimization/hillclimbing/hill-climbing-helpers.metta
K1rubel: Me neither!

optimization/hillclimbing/hill-climbing-helpers.metta
Yagth: Yeah but I was thinking something to add the floor-math after the foldl-atom finishes not during. So the final code looks like:
```scheme

(= (informationTheoreticBits $knobMap)
   (floor-math (foldl-atom (MultiMap.values $knobMap) 0 $bits $knob
                (+ $bits (knobInformationBits $knob))))
)
```

deme/tests/expand-demes-test.metta
```diff
@@ -175,36 +184,75 @@
...
-    ($_ (println! $optimizeDemesReturn))
-    (($instance (mkDeme $rep (mkSInstSet $instSet) $id) $state) $optimizeDemesReturn)
+    (($instance (mkDeme $rep (mkSInstSet $instSet) $id) $state) $optimizeDemeReturn)
     ($instanceCount (List.length $instSet))
     ($scoredElems (List.map isScored $instSet))
 )
 ((>= $instanceCount 4) (List.any $scoredElems)))
 (True True))
 
-;; ;; Testcase for expandDeme INFO: For some reason this test case disturbs the next test case but is fine alone.
```
Yagth, line 196: I wonder why this was removed 🤔

optimization/hillclimbing/hill-climbing-helpers.metta
```diff
@@ -372,23 +666,32 @@
...
         ;; Termination check compares against the scalar `$bestPossibleScore`,
         ;; which carries the *raw* best score (matching classic MOSES). So we
         ;; must extract the raw-score field from the candidate's Cscore here,
         ;; not the penalty-adjusted one -- comparing penScore against a raw
         ;; target would mix the complexity penalty into the threshold.
-        ($scoreToCompareBelow (if (or (== $rep ()) (== $id ())) $newBestSscore (getScore $newBestSscore)))
+        ($scoreToCompareBelow
+          (if (or (== $rep ()) (== $id ()))
+              $newBestSscore
+              (bestRawDeme $newDeme)))
```
Yagth, line 688: Isn't tracking the bestRaw score better than re calculating the best score every time?

optimization/hillclimbing/test/hillclimbing-helpers-test.metta
Yagth: But does our generateNeighborhood match this pattern? I think the classic implementation must be generating instances at exactly d distances as well instead of up to d positions just like the estimate as well? Do you think we have to update that as well?

moses/demo-problems.metta
```diff
@@ -256,7 +248,15 @@
       )
       ()))
 
+;; Dispatch the configured problem to its domain-specific setup; both setup
+;; paths subsequently enter the same shared runMoses search lifecycle.
 (= (moses-run)
+   (case (param problem)
+      ((ant (moses-ant-run))
```
Yagth, line 255: This ant run is specific to the ant experiment right?

metapopulation/exemplar-selection.metta
```diff
@@ -28,6 +31,71 @@
...
+   (selectExemplar (eligibleExemplars $metaPop $visits $revisit)))
+
+;; Fold raw fitness without allocating a score list. Starting from the first
+;; row preserves the non-empty-input contract. C++ prefers lower complexity on
+;; a raw-score tie; an exact tie keeps the earlier exemplar.
+;;
+;; Directly destructure both exemplars here. Although getExemplarCscore and
+;; getComp work independently, PeTTa currently fails to compose those accessor
+;; calls inside foldl-atom and silently returns no result for a multi-row list.
+(= (bestRawExemplar ()) ())
```
Yagth, line 73: Is this used anywhere?

deme/expand-deme.metta
```diff
@@ -1,159 +1,399 @@
...
+;; termination, and finalization for every problem family.
+(= (runMoses $maxGen $maxScore $metaPop $nExpansion $nDeme
+             $context $optimize)
    (let* (
-          ($knobs (Map.values $dscMp))
-          ($defaults (List.map getKnobDefault $knobs))
-          ($finalDefaults (List.map second $defaults))
+          ($_reset (resetCandidateScoreCache))
+          ($_domainReset (resetDomainRunCaches $context))
+          ($_seed (seedInitialCandidateCache $context $metaPop))
```
Yagth, line 252: We  don't call runMoses more than once do we? Why is this reset necessary here?

Yagth (approved):

optimization/hillclimbing/hill-climbing-helpers.metta
K1rubel: That we can do!

optimization/hillclimbing/hill-climbing-helpers.metta
```diff
@@ -150,28 +186,210 @@
...
+     (((mkStrategyScoreContext $nGames $complexityRatio)
+       (getCandidate (mkNodeId (0)) $rep $id $inst))
+      ($other (getCandidate $rep $id $inst)))))
+
+;; Score a materialized program without reconstructing it through the
+;; representation-specific scorer.
+(= (scoreMaterializedTree
+      $tree
+      (mkAntContext $steps $complexityRatio $actions $perceptions))
+   (getAntCscore $tree $complexityRatio $steps))
```
Yagth, line 237: So did we define this function somewhere? I remember the definition wasn't in the PR last time I reviewed?

optimization/hillclimbing/hill-climbing-helpers.metta
K1rubel: Will devise a better approach, one that doesn't require scanning the whole deme everytime.

optimization/hillclimbing/hill-climbing-helpers.metta
K1rubel: This is introduced as part of the integration attemp

metapopulation/exemplar-selection.metta
K1rubel: Was used in `expand-deme.metta` line 263

deme/expand-deme.metta
K1rubel: Sure that's right! For a guaranteed single run the first two resets are not necessary. The third reset is necessary to construct the initial exemplar in the santaFe problem Moses run

Yagth:
Merging this for now. Separate PR is better for the improvements any way :)

